PR #14, opened 2015-10-19
Added support for HTML forms (fixes #11)

### LINKED ISSUE #11: Support HTML forms
- Use the approach as Polymer elements (iron-form)
- Implement `Polymer.IronFormElementBehavior`
- `iron-form` could be used in testing (`serialize()` maybe?)

tomivirkki:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/14)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,6 +22,8 @@
     "components-demo-resources": "vaadin/components-demo-resources#master",
     "iron-dropdown": "polymerelements/iron-dropdown#~1.0.5",
     "iron-flex-layout": "PolymerElements/iron-flex-layout#~1.0.3",
+    "iron-form": "polymerelements/iron-form#~1.0.10",
+    "iron-form-element-behavior": "polymerelements/iron-form-element-behavior#~1.0.4",
     "iron-icon": "PolymerElements/iron-icon#~1.0.4",
     "iron-icons": "PolymerElements/iron-icons#~1.0.3",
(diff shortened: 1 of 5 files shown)


Saulis:
I think it should have an example of using inside `iron-form` otherwise :+1: 


---

Reviewed 4 of 4 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/14)

tomivirkki:
Added an example


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/14)

jounik:
So you always need to use `iron-form` for it to work, right?

tomivirkki:
Vaadin-combo-box local DOM does contain an <input> that's decorated with the 'name' and 'required' properties as attributes so it can be used in a plain HTML form also; as long as native shadow DOM isn't enabled. That's the case with <paper-input> also.

Note: There was some issues with this PR regarding the use in and <iron-form>, fixed in https://github.com/vaadin/vaadin-combo-box/pull/18

jounik:
So if Iʼm using native shadow DOM, does it work with iron-form at least?

tomivirkki:
Sure does. The shadow dom limitation only concerns use in a plain HTML form (non iron-form)

Issue #16, opened 2015-10-20
Keyboard navigation


Saulis:
Features / Behavior:
 -  [x]  Let's focus on desktop behavior, if it works on fullscreen, great - if not then not.
 - [x] Input can be focused using TAB
 - [x]  Popup opens when you start typing on it (or click the icon)
 - [x]  Popup opens when focused and you push DOWN/UP
 - [x]  If a value is selected when popup opens, it's highlighted, otherwise nothing is highlighted. Selected item is not highlighted after filtering anymore.
 - [x]  After typing in a filter, the topmost item is highlighted
 - [x]  Focus/highlight is moved up and down by pushing DOWN/UP
 - [x]  When you navigate to the topmost item, and push one more time UP, the topmost item stays focused??
 - [x]  You need to be able to navigate the value inside the input field using LEFT/RIGHT
 - [x]  Pushing ENTER will select the focused/highlighted item
 - [ ]  When the input box is cleared and ENTER is pushed, combobox value is cleared.
   - Q: How is the input box cleared ? A: With BACKSPACE
   - Q: When  input box is empty is it cleared ? A: combobox.value is cleared if ENTER is 
hit when input box is empty/cleared
   - Additional info: when input value is changed, the items are reset and item highlight is removed.
 - [ ]  When the input box is cleared we don't want to highlight the topmost item
 - [x]  Pushing SPACEBAR is for now reserved for multi selection
 - [x]  Pushing ESCAPE will cancel and close the popup

Design questions:
- When you navigate to the topmost item, and push one more time UP, the topmost item stays focused? or should it remove focus

Saulis:
One feature missing from our list is that the list of items should of course scroll when it's navigated down or up. I think that we need to implement that in #22 though.

Saulis:
@jounik @tuomasnironen have we discussed if the overlay should be modal or not?

If not, then we can set `allowOutsideScroll` `true` in `<iron-dropdown>` to receive arrow down and up events correctly. Otherwise we need to have a monkey patch to [hack into the dropdown scrolling manager.](https://github.com/vaadin/vaadin-combo-box/commit/7d178b5319ff0c99282c9171ce3a31a7a3651000#diff-f8444860963b469ee2b3ef1765ecb354R119)

Saulis:
[...]

PR #20, opened 2015-10-22
Feature/keyboard

manolo:
Connected to #16

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/20)


tomivirkki:
iron-a11y-keys-behavior might be useful here (example: https://github.com/PolymerElements/iron-menu-behavior/blob/master/iron-menu-behavior.html#L63)

PR #71, opened 2015-12-03
fire change-selected event

JITBALJINDER:
fire the event `change-selected` when the selected item changes

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/71)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -399,6 +399,7 @@
       var _val = this.$.selector.selectedItem;
       if (_val !== null) {
         this.value = _val;
+        this.fire('change-selected', this.value);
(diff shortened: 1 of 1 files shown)


tomivirkki:
Thanks for the PR. Vaadin-combo-box uses iron-form-element-behavior which marks value-property as [notified](https://github.com/PolymerElements/iron-form-element-behavior/blob/master/iron-form-element-behavior.html#L47) so whenever it changes, vaadin-combo-box automatically fires a "value-changed" event.

See this in action at https://cdn.vaadin.com/vaadin-elements/master/vaadin-combo-box/demo/#toc_1_selecting_a_value

PR #90, opened 2015-12-14
Add a viewport threshold on full screen mode

### LINKED ISSUE #23: Add a viewport size threshold for fullscreen mode
The fullscreen mode should not depend on any user agent detection. The viewport size is the only thing that should trigger it (but it does not need to adapt when the item display is open).

The threshold should detect both viewport width and viewport height, and take the smallest of those. If either is under 600px, then use fullscreen.

platosha:
Resolves #23.

From now and then the overlay opens in full screen mode when the viewport
width is below the `responsiveWidth` property value (600px by default). The
viewport width check is performed each time the overlay opens.

The `fullScreen` overlay’s property changed to be read-only. The full
screen mode can be turned on manually by setting the overlay’s
`forceFullScreen` property to `true`.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/90)

### DIFF
--- a/test/adaptive-overlay.html
+++ b/test/adaptive-overlay.html
@@ -42,9 +42,14 @@
       describe('fullscreen', function() {
 
         beforeEach(function(done) {
-          overlay.fullScreen = true;
+          overlay.responsiveWidth = '600px';
+          overlay.forceFullScreen = true;
           overlay.open();
           Polymer.Base.async(done);
+        });
+
+        it('should open in fullscreen mode when forced', function() {
+          expect(overlay.fullScreen).to.be.true;
         });
 
         it('should have correct dimensions', function() {
@@ -89,9 +94,14 @@
       describe('non-fullscreen', function() {
 
         beforeEach(function(done) {
-          overlay.fullScreen = false;
+          overlay.responsiveWidth = '0px';
+          overlay.forceFullScreen = false;
           overlay.open();
           Polymer.Base.async(done);
+        });
+
+        it('should not open in fullscreen mode', function() {
+          expect(overlay.fullScreen).to.be.false;
         });
 
         it('should have correct target elements', function() {
@@ -140,6 +150,33 @@
           overlay.open();
         });
 
+      });
+
+    });
+
+    describe('responsiveWidth property', function() {
+      var overlay;
+
+      beforeEach(function(done) {
+        overlay = fixture('overlay').firstElementChild;
+      });
+
+      it('should open in fullscreen mode if responsiveWidth is >= viewport', function(done) {
+        overlay.responsiveWidth = window.innerWidth + 'px';
+        overlay.open();
+        Polymer.Base.async(function() {
+          expect(overlay.fullScreen).to.be.true;
+          done();
+        });
+      });
+
+      it('should open in non-fullscreen mode if responsiveWidth is < viewport', function(done) {
+        overlay.responsiveWidth = (window.innerWidth - 1) + 'px';
+        overlay.open();
+        Polymer.Base.async(function() {
+          expect(overlay.fullScreen).to.be.false;
+          done();
+        });
       });
 
     });
--- a/test/adaptive-overlay.html
+++ b/test/adaptive-overlay.html
@@ -119,10 +119,12 @@
 
           overlay.parentElement.style.float = 'left'; // disables margins collapsing
 
-          overlay.open();
+          var interval = setInterval(function() {
+            if (overlay.$.dropdown._openChangedAsync) {
+              return;
+            }
 
-          Polymer.Base.async(function() {
-            var target = overlay.$.dropdown._verticalAlignTargetValue;
+            var target = overlay.$.dropdown.getBoundingClientRect().top;
 
             // When multiplied by `devicePixelRatio`, target should be integer
             expect(target * (window.devicePixelRatio || 1) % 1).to.be.equal(0);
@@ -131,8 +133,11 @@
             overlay.parentElement.style.removeProperty('margin-top');
             overlay.parentElement.style.removeProperty('float');
 
+            clearInterval(interval);
             done();
-          });
+          }, 100);
+
+          overlay.open();
         });
 
       });
--- a/test/common.js
+++ b/test/common.js
@@ -8,6 +8,9 @@ var fullScreen = (function() {
   }
 })();
 
+// Set the test runner window width to trigger fullscreen mode
+window.parent.document.querySelector('#subsuites').style.width = fullScreen ? '500px' : '700px';
+
 var describeSkipIf = function(bool, title, callback) {
   bool = typeof bool == 'function' ? bool() : bool;
   if (bool) {
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -52,7 +52,7 @@
     [false, true].forEach(function(isFullScreen) {
       describe('setting the input field value in ' + (isFullScreen === true ? 'fullscreen' : 'desktop'), function() {
         beforeEach(function() {
-          comboBox.$.overlay.fullScreen = isFullScreen;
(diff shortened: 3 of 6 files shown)


jounik:
Didn’t we decide it wasn’t necessary to have a separate ”force fullscreen”
flag/property, as the same can be done by setting the responsiveWidth
property to 0px?
ma 14. joulukuuta 2015 klo 17.12 Anton Platonov <notifications@github.com>
kirjoitti:

platosha:
To be precise, the same can be done by setting `responsiveWidth` property to some big number like `32767px`. Yes, we decided. I made these properties after `paper-drawer-panel`, it has `responsiveWidth` and `forceNarrow`. Thought it would be better, more convenient and consistent this way.

The `fullScreen` property is no more suitable for direct assignment, so `forceFullScreen` could be used instead. It is also a boolean flag, that’s convenient.

The difficult part of this PR is handling side-effects of switching modes, but not adding the property itself.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5ZVYhL6S-_NvabeEp4:660978)*

tomivirkki:
If forceFullScreen is added then shouldn't forceNonFullScreen also be added to make the API coherent? Then what would happen if they're both set true? If it's for internal purposes only then it should start with an underscore _forceFullScreen

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5ZZyvcWy5QL37E7gaX:-1699410428)*

platosha:
No, it shouldn’t. You can force non-fullscreen mode on any viewport by setting the `responsiveWidth` to `0`.

When `forceFullScreen` is set to true, then there’s no width check, i. e. it overrides the `responsiveWidth`.

I still think this is convenient for both internal and external use.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5ZdHr9wLRUSmLCPSg3:1775136039)*

tomivirkki:
On iPad the dropdown always opens below the component, even if vaadin-combo-box is at the bottom of viewport

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-adaptive-overlay.html, line 386 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Ze2nFYggWpY0Lyy9s:-K5Ze2nFYggWpY0Lyy9t:-1268542774) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L386)):*
These should be moved to CSS

#dropdown[ios]:not([fullscreen]) {
  left: 0 !important;
  top: 100% !important;
}

#dropdown[vertical-align='bottom'][ios]:not([fullscreen]) {
  top: 0% !important;
  bottom: 100% !important;
}

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5Zh0OoGtzvW9eJqjyK:793209120)*

Saulis:
@tomivirkki it's because the iPad will scroll the input field upwards - there's always going to be more space below than above.

@platosha as discussed, there's something fishy when opening the dropdown on iPad by clicking the input field - no items are rendered at all in some cases. (scrolling triggers re-rendering)

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-adaptive-overlay.html, line 349 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Z_iGfhtoXwUsjlVWy:-K5Z_iGfhtoXwUsjlVWz:-1021577488) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L349)):*
I think we need a comment for `!this.ios` - e.g. 'iOS scrolls the focused inputs into view, so there's always more space below the input.' Which makes me think that we'll be in trouble when using this overlay component without an input field, but let's worry that later :smile: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5ZnLIXF5lBJ7pHOTm8:577028081)*

platosha:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-adaptive-overlay.html, line 349 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Z_iGfhtoXwUsjlVWy:-K5Zq3pb3DpC_KfitGtM:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L349)):*
Done.

---

*[vaadin-adaptive-overlay.html, line 386 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Ze2nFYggWpY0Lyy9s:-K5ZpOJSeKNyZ_IGyr--:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L386)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90)*

tomivirkki:
*[vaadin-adaptive-overlay.html, line 274 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Zyk7r2StXjBbX3JeN:-K5Zyk7r2StXjBbX3JeO:-569251140) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L274)):*
Move this logic to _computeFullScreen-function instead and make fullScreen a computed property
computed: '_computeFullScreen(opened)'

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90)*

platosha:
The iPad and Android issues discovered by @saulis are resloved.

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-adaptive-overlay.html, line 274 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-K5Zyk7r2StXjBbX3JeN:-K5a84l07zTDvHpgvsMx:2016331132) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/a243aa8ad583580457e8f40b3fd149f751f2548b/vaadin-adaptive-overlay.html#L274)):*
Just have tried that. This is not a good idea. The current approach seems more consistent for me.

There is some clean up code in `_closed`, which depends on if the overlay was previously opened in fullscreen mode or not. If we rewrite the `fullScreen` property to be computed depending on `opened`, it would recalculate before `_closed`, right after `opened` changes to `false`. So we would need to introduce some other property to store the `fullScreen` value at the moment when the overlay opens. Additional complexity for no reason.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/90#-:-K5a9gXIrnla2M0lAmy7:-2071613052)*

Saulis:
Will be replaced by changes done in #99

PR #99, opened 2015-12-21
Enable body scrolling.

### LINKED ISSUE #98: Enable scrolling when overlay is open
`<iron-dropdown>` disables scrolling by default when the overlay is open. 

If the body would be scrollable instead, it would allow user to scroll hidden items into view more easily.

Update: decided to get rid of `<iron-dropdown>` altogether to:
 - Enable scrolling by default and remove the need for a fullscreen mode and vertical alignment
 - Use `position: absolute` for the overlay to mitigate iOS Safari issues
 - Place the overlay directly under body to handle `z-index` and scrolling containers

Saulis:
Fixes #98

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/99)

### DIFF
--- a/test/adaptive-overlay.html
+++ b/test/adaptive-overlay.html
@@ -9,6 +9,7 @@
   <script src="../../test-fixture/test-fixture-mocha.js"></script>
 
   <link rel="import" href="common.html">
+  <script src="../../promise-polyfill/Promise.js"></script>
   <script src="common.js"></script>
 </head>
 
@@ -50,26 +51,27 @@
 
       describe('fullscreen', function() {
 
-        beforeEach(function(done) {
+        beforeEach(function() {
           overlay.fullScreen = true;
           overlay.open();
-          Polymer.Base.async(done);
+
+          return waitUntilOpen(overlay);
         });
 
         it('should have correct dimensions', function() {
           var rect = overlay.$.dropdown.getBoundingClientRect();
-          expect(rect.left).to.equal(0);
-          expect(rect.top).to.equal(0);
+          expect(rect.left).to.equal(0, 'left');
+          expect(rect.top).to.equal(0, 'top');
           expect(dropdownContent.clientWidth).to.equal(document.documentElement.clientWidth);
           expect(dropdownContent.clientHeight).to.equal(document.documentElement.clientHeight);
         });
 
-        it('should alter the viewport info', function(done) {
+        it('should alter the viewport info', function() {
           expect(viewportInfo.content).to.equal(overlay._viewportInfo._fullScreenContent);
           overlay.close();
-          Polymer.Base.async(function() {
+
+          return waitUntilClosed(overlay).then(function() {
             expect(viewportInfo.content).not.to.equal(overlay._viewportInfo._fullScreenContent);
-            done();
           });
         });
 
@@ -97,10 +99,11 @@
 
       describe('non-fullscreen', function() {
 
-        beforeEach(function(done) {
+        beforeEach(function() {
           overlay.fullScreen = false;
           overlay.open();
-          Polymer.Base.async(done);
+
+          return waitUntilOpen(overlay);
         });
 
         it('should have correct target elements', function() {
@@ -152,12 +155,13 @@
       });
 
       describe('scrolling in fullscreen mode', function() {
-          beforeEach(function(done) {
+          beforeEach(function() {
             overlay = fixture('scrollable-overlay').firstElementChild;
             overlay.fullScreen = true;
 
             overlay.open();
-            Polymer.Base.async(done);
+
+            return waitUntilOpen(overlay);
           });
 
           it('should disallow outside scroll', function() {
@@ -189,12 +193,13 @@
           return e;
         }
 
-        beforeEach(function(done) {
+        beforeEach(function() {
           overlay = fixture('scrollable-overlay').firstElementChild;
           overlay.fullScreen = false;
 
           overlay.open();
-          Polymer.Base.async(done);
+
+          return waitUntilOpen(overlay);
         });
 
         it('should allow outside scroll', function() {
--- a/test/adaptive-overlay.html
+++ b/test/adaptive-overlay.html
@@ -24,6 +24,15 @@
     </template>
   </test-fixture>
 
+  <test-fixture id="scrollable-overlay">
+    <template>
+      <div style="width: 100px; height: 100px;">
+        <vaadin-adaptive-overlay>
+          <div class="content" style="height:10000px">content</div>
+        </vaadin-adaptive-overlay>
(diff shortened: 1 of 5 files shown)


manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/adaptive-overlay.html, line 123 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/99#-K63h9yYeomahRo6IU-_:-K63h9yZWmalm2mh0T4G:-1605896818) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/56e5e7824d98811bde0d31ec1c5b72de018d1832/test/adaptive-overlay.html#L123)):*
Can this be moved out of the timer, so as the timer is not fired if the _openChangedAsync has not changed?

---

*[test/adaptive-overlay.html, line 138 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/99#-K63hS5tavySLs5Ufa7F:-K63hS5tavySLs5Ufa7G:1160793073) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/56e5e7824d98811bde0d31ec1c5b72de018d1832/test/adaptive-overlay.html#L138)):*
Why do we need a timeout ?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/99)*

Saulis:
It's been a LONG day for me and @platosha :smile: 

FF is still not working when vertical alignment is bottom because of iron-fit-behavior bug.

We should maybe reconsider just getting rid of the vertical aligment feature at this point, open the overlay always below the input and having a constant height for the overlay.

Saulis:
I've added a summary to us track the progress after we get back to work from holidays.

Demo/prototype 0  is built on the PR version, and it works as intended, after many hours of manual testing the FF issue mentioned is the only thing that remains. Prototypes 1-3 are built on top of each other and they try to tackle each others shortcomings.

Would :heart: to hear your comments on these @jounik .

http://wc.demo.vaadin.com/combo-box-scrolling-enabled/
0: Base version with scrolling enabled
 - Enabled `allow-outside-scroll` on `<iron-dropdown>`
 - Added 8px margins for the overlay to clearly visualize where the edge is.
 - Prevented kinetic scrolling energy from bubbling outside the overlay.
 - After enabling scrolling, https://github.com/PolymerElements/iron-fit-behavior/issues/26 became quite obvious with FF - overlay height is set incorrectly and you can't scroll the iron-list.
   (although it's still better than in master with FF)

http://wc.demo.vaadin.com/combo-box-scrolling-max-height-65vh/
1: 0 -> Have a max-height for the overlay
  - Any height value works, but `65vh` felt quite nice. 
    Anything close to 100vh makes it really hard to fit to the viewport by scrolling, too small values
    don't show enough items on the list.
  - Pro: Having max-height makes the overlay usable in FF
  - Con: Don't know why, but max-height prevents the overlay from aligning with the bottom or top and makes it always to have a height of max-height.
  - Con: Sometimes, the overlay is only partly visible when opening, forcing the user to scroll the body.
  - Con: As it is currently, keyboard navigation doesn't reveal items highlighted outside viewport.
  
http://wc.demo.vaadin.com/combo-box-scrolling-vertical-align-frozen/  
2: 1 -> Open the overlay always below input
  - Vertical alignment feature was originally added because of the limited scrolling - now that the body  can be scrolled, this feature might not be needed anymore.
  - Pro(??): Opening dropdown always below the input is more natural?
  - Pro: Keyboard navigation was somewhat unintuitive when overlay was opening above the input
  - Pro: Simplifies the component, makes it use less features from `iron-dropdown`.
  - Con: User is forced to scroll the body to see all the items more often now.

http://wc.demo.vaadin.com/combo-box-scrolling-yo-dawg-scrolling/
3: 2 -> Scroll the body automatically after opening the overlay
  - Don't know if the users will like it or hate it, we could of course have it as a configurable parameter
  - Pro: Eliminates the need for a user to scroll
  - Pro: Enables using larger max-heights for maximum visibility
  - Con: Not sure how nicely this plays with touch devices which scroll when input field is under the virtual keyboard.

Saulis:
Just realized that if the scrolling animation in Prototype 2 is done right, we might be able to use it to 'override' the iOS scrolling after input focus. This way we could always position the main input at the top of the viewport in 'fullscreen' mode and get rid of the secondary input. And also most likely some async repositioning logic.

Saulis:
Update: made a few 5 minute prototypes more for the fullscreen.

http://wc.demo.vaadin.com/combo-box-scrolling-no-fullscreen/
4: 3 -> "Fullscreen" disabled for all devices
  - Has exactly the same styles and settings as "desktop" mode
  - Manual scrolling when opening seems to work pretty nicely on iPhone at least.
  - We still need an attribute to detect touch devices to disable the item :hover styles for example for touch devices.
  - Pro: Simplifies the component even more
  - Pro: Removes the secondary input
  - Con: We might encounter some styling issues later

http://wc.demo.vaadin.com/combo-box-scrolling-fullscreen-tweaks/
5: 4 -> Added minor tweaks for "fullscreen" devices
  - Prototype has a threshold width of 450px below which it goes to fullscreen mode (might need page refresh when changing phone orientation)
  - Smaller margin on top
  - Body scrolling disabled when overlay open.
  - I did test also to set the overlay size to fit the whole screen and added a resizing animation - worked in theory, but left it out of the prototype because it wasn't polished enough.

manolo:
- The autoscrolling (3:2) is upsetting in desktop, for mobile is a convenient feature though.

- Touching the glass layer to close the drop box is alright, but other iterations like swipe should work as well.

IMO, the UX not having the fullscreen overlay is more natural, the UI is always the same in all devices and it's very usable

manolo:
LGTM

---

Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[test/common.js, line 43 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/99#-K7q3-y3m9oSsr-LJGQp:-K7q3-y4MxbTnPH-z61o:687793577) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/8a42f9a4f07050f742b0cdc58ab4f8f5c6a33ddd/test/common.js#L43)):*
I like more the syntax you use below : window.Promise = window.Promise || MakePromise

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/99#-:-K7q3aWinbAkEfCHjsRR:676834233)*

Saulis:
Work has been moved to #105.

PR #115, opened 2016-01-26
Prevent label from overflowing and causing zoom flickering

### LINKED ISSUE #114: Selecting a value or typing to the input causes zoom flickering on iOS
Selecting a value or typing to the input causes zoom flickering on iOS like illustrated in the attached Gif. The blue background for the `label` is added intentionally to demonstrate the actual cause of the issue.

zoom-issue-ios

tehapo:
Fixes #114

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/115)


tomivirkki:
Does this also fix the same issue? https://github.com/PolymerElements/paper-input/commit/f5e467b9fcec5931ab1759fbc0a8763169c7252b

If so, we could just upgrade the dependency

PR #120, opened 2016-02-01
Do not force version in dependencies.

### LINKED ISSUE #119: vaadin-combo-box forces to install old versions of iron components
It would be nice if it depends on non specific versions to avoid conflicts.

Related with vaadin/gwt-polymer-elements#93

manolo:
Fixes #119 vaadin/gwt-polymer-elements#93

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/120)


Saulis:
Were also considering doing this, but in a bit more "controlled" manner by first enabling the screenshot tests and after that changing the dependencies. For example the `paper-input` was recently updated to 1.1.3 because we noticed that an older version has visual issues on iOS. So we might want to "force" some dependencies to be newer than 1.0.0.

So I wouldn't merge this without some visual regression tests.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/120#-:-K9R5giauVp2FlLrlCjO:307042896)*

Saulis:
Discussed with Jouni, I'll prepare a visual test plan and test everything after updating the dependencies to we can update the dependencies. We also discussed that we should loosen the dependency restrictions, but we can't reliably state that every version since 1.0.0 will work so we should use the current lower limits we have.

manolo:
Ok, I will update paper-input to be greater or equal to the min version we need, but I think visual tests should be in a different PR and shouldn't block this.

Saulis:
Added a visual test plan, which I used to manually test everything on Chrome, FF, Safari, IE11, Edge, iOS and Android. Found some issues with Edge and made separate tickets out of those. :shipit: 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/120#-:-K9g_OScXWrahp6kmsXw:-1370457220)*

Issue #122, opened 2016-02-02
Overlay is not closed on blur if focus is moved to document.body


Saulis:
Overlay doesn't close if focus is moved to for example browser toolbar, which makes `document.activeElement` getting changed to `document.body`

Saulis:
Had to revert this in 29c2a1e due to regressions which made the input blur close the overlay when clicking on items and toggle button. 

Work continues in #132

tomivirkki:
Works with the latest version. Closing.

PR #132, opened 2016-02-04
Focus overlay explicitly when clicked on.

Saulis:
We are effectively blurring the input field when overlay is tapped in order to
hide the virtual keyboard on touch devices.

However, at least on WIN10 IE11 and Edge the focus is moved to document.body
when the input is blurred, forcing the overlay to be closed and selected value
to be reverted.

Setting the focus to overlay (#scroller) now that it has tabindex fixes this,
and also makes the code reflect better on what we're trying to achieve.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/132)

### DIFF
--- a/test/selecting-items.html
+++ b/test/selecting-items.html
@@ -76,6 +76,14 @@
       combobox.$.overlay.$.selector.toggleSelectionForItem('foo');
       expect(valueChangedSpy.callCount).to.equal(0);
     });
+
+    it('should focus on overlay when items are clicked on', function() {
+      combobox.open();
+
+      var item = combobox.$.overlay.fire('down');
+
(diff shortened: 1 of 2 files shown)


Saulis:
Needs some more work, focusing explicitly on scroller makes the window scroll. Also, need to make toggle and clear icons focusable so that they don't close the overlay (currently happening on master)

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/132#-:-K9ggG3j9ufYLtP89NIH:-341648775)*

Saulis:
29c2a1e reverted to fix this issue, #122 reopened

Issue #129, opened 2016-02-04
Underline is misplaced on hover in MS Edge when using prefixes and suffixes


Saulis:
Initial state:

![image](https://cloud.githubusercontent.com/assets/1039468/12812609/a79866c0-cb3b-11e5-84a4-7710209424a4.png)

After hover (removing hover doesn't return initial state):

![image](https://cloud.githubusercontent.com/assets/1039468/12812623/b6703498-cb3b-11e5-8e53-77051cf4111e.png)

Focus + after hover:

![image](https://cloud.githubusercontent.com/assets/1039468/12812652/dd2c170a-cb3b-11e5-9469-049c9fcc3164.png)

Also, similar kind of "double" underline appears under the toggle and clear icon buttons when clicking on them.

Was able to see the same behavior in Polymer Demos: https://elements.polymer-project.org/elements/paper-input?view=demo:demo/index.html&active=paper-input (v1.1.5), but was not able to produce a JSBin for a bug report to paper-input.

Saulis:
Related with polymerelements/paper-input#329

Saulis:
Related with polymerelements/paper-input#329

notwaldorf:
I've been able to repro this with a `paper-input`, but it only seems to happen (for me) if the `paper-input` is inside a form: https://jsbin.com/roqudo/edit?html,output. The weirdo underline only appears between the prefix and the suffix, which means a style is being applied to the native `<input>` itself.

I'll try to come up with a much simpler example tomorrow (just an `<input>` with an underline), because I'm suspicious this is an Edge rendering bug (or, best case scenario, a weird CSS style)

notwaldorf:
Annnnd I just tried that demo on an actual Edge browser and not a VM (20.10240.16384.0) and I don't see any artifacts at all, so :fire: everything is fine :fire:

Saulis:
@notwaldorf thanks for looking into this. The screenshots were taken in a VM running on OS X with Retina display, but IIRC I had some problems reproducing the problem on a Surface Pro tablet, so it _might_ be a resolution related issue.

@samiheikki has been looking into this and he is using a native WIN10 + Edge and he has been able to reproduce the issue at least under certain conditions.

Update: I'm not 100% sure anymore if I took the screenshots on my laptop screen or on my non-retina external display, I need to double-check that tomorrow.

notwaldorf:
[...]

PR #137, opened 2016-02-05
Fix for input value bottom cutting off in IE11

### LINKED ISSUE #130: Input value bottom gets cut off in IE11
For values with letters like `y`, the bottom half gets cut off. Seems to happen only on IE11 (tested on Windows 7)

image

samiheikki:
Fixes #130

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/137)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -103,6 +103,10 @@ enable usage within an `iron-form`.
       display: none;
     }
 
+    #input {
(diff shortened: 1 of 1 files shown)


Saulis:
Seems to also help partly for #129, removing the problem when hovering. When keyboard navigating, the  issue seems to still remain.

PR #143, opened 2016-02-10
Added disabled property to combobox.

jermsv1:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/143)


Saulis:
Hi there! If you would be kind enough to sign the CLA, we would be happy to take this PR in.

jounik:
Don’t mean to be a party pooper, but we will need to add tests for this also.

Does the element already prevent the dropdown from opening when the element is disabled?

A bit off topic, but it seems weird that paper-input still allows you to focus the field when it’s disabled. I would expect to be able to skip the field with the tab key in a form if it is disabled.

jounik:
Forgot to say – like Sauli mentioned – thank you for the PR, we’d be happy to take it!

vaadin-dropdown-behavior.html
```diff
@@ -119,4 +124,4 @@
       }
     }
   };
-</script>
+</script>
```
jounik: Minor nitpick: unnecessary white space change.

Saulis:
@samiheikki and myself will be helping out by adding tests and other missing pieces.

jermsv1:
Thank you! I'm a little tied up with another project and won't get to it
till later next week

Saulis:
We've moved the changes to PR #149

Issue #91, opened 2015-12-14
Unify sizing with paper-inputs


vaadin-marcus:
Currently `vaadin-combo-box` looks a bit out of place in a form with other paper-inputs as it is sized or aligned differently.

<img width="811" alt="screen shot 2015-12-14 at 8 54 33 am" src="https://cloud.githubusercontent.com/assets/883698/11787620/88217c50-a241-11e5-9fff-2fbf031c3cfb.png">

jounik:
1d timebox for the research

manolo:
Probably fixed with : https://github.com/vaadin/vaadin-combo-box/pull/95

PR #144, opened 2016-02-11
Fix for input value bottom cutting off in IE11

### LINKED ISSUE #130: Input value bottom gets cut off in IE11

samiheikki:
Fixes #130

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/144)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -58,6 +58,10 @@ enable usage within an `iron-form`.
     paper-input-container {
       position: relative;
       padding-bottom: 0;
+
+      --paper-input-container-label-floating: {
+        line-height: 24px !important;
+      };
     }
 
     paper-input-container[opened] {
@@ -103,6 +107,12 @@ enable usage within an `iron-form`.
       display: none;
     }
 
+    #input {
+      line-height: normal;
+    }
+    #label {
+      line-height: inherit;
+    }
   </style>
 
   <template>


jounik:
Just checking, that I understand this right: this only happens in vaadin-combo-box, and not in paper-input? Just wondering why we need a patch for it in our element, when it seems like it’s caused by paper-input or paper-input-container.

Saulis:
At least it's not happening in the `paper-input` demos on IE11. ..

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/144#-:-KAZ2Q7TTrIe990ivPQ8:1025113059)*

platosha:
Works for me.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/144#-:-KA_miXWmFK17Y-1_T8M:-698977764)*

jounik:
This still seems like it’s not the right fix, as it again causes #91 to not be okay. There are minor alignment issues in Safari, Chrome and Firefox (click the screenshot for full view).

combo-box-vs-paper-input

Messing with the internal styles of paper-input-container feels like we’re fixing it from the wrong place.

Also, this `padding-bottom: 0;` is causing vertical-alignment issues in IE11 at least, and to me feels totally wrong to override: https://github.com/vaadin/vaadin-combo-box/blob/master/vaadin-combo-box.html#L60

The padding affects the dropdown position, yes, but we should fix that somehow else, maybe by computing the position from some other element, or then computing the bottom padding and taking that into account.

jounik:
It actually seems like an issue in paper-input also. Win 8.1, IE11:

screen shot 2016-02-19 at 11 11 06

jounik:
So I would want us to remove this fix. I’ll create a separate issue about the bottom padding override.

tomivirkki:
Link to the paper-input issue https://github.com/PolymerElements/paper-input/issues/171

jounik:
Thanks Tomi, just searched for it myself as well and found the same thing :)

Saulis:
I'll revert the commit :smiley_cat:

PR #172, opened 2016-02-25
Language corrections

jojule:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/172)

### DIFF
--- a/docs/vaadin-combo-box-basic.adoc
+++ b/docs/vaadin-combo-box-basic.adoc
@@ -27,9 +27,8 @@ combobox.value = 'Bromine';
 
 == Events
 
-The [vaadinelement]#vaadin-combo-box# element dispatches events when the value changes.
+The [vaadinelement]#vaadin-combo-box# element dispatches events when the selection value changes.
 An event contains the [propertyname]#value# property in the [propertyname]#event.detail# object.
-
 
 [source,javascript]
 ----
@@ -64,7 +63,7 @@ When using the element inside another Polymer element, you can use the declarati
 
 == Forms
 
-In addition to other features mentioned earlier, the [vaadinelement]#vaadin-combo-box# implements the [classname]#IronFormElementBehavior#, so it can be used as a field in an [elementname]#iron-form#.
+The [vaadinelement]#vaadin-combo-box# implements the [classname]#IronFormElementBehavior#, so it can be used as a field in an [elementname]#iron-form#.
 
 [source,html]
 ----


jounik:
@magi42, please review the second change (adding "The" in front of vaadin-combo-box).

For the "selection value", to me it seems misleading. There are two properties, "value" and "selectedItem", but only the value fires an event when it changes. Not sure if we should mix them in the documentation.

magi42:
Reviewed 1 of 1 files at r1.

---

*[docs/vaadin-combo-box-basic.adoc, line 66 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/172#-KBRgEjI1EsbedV7ZF-t:-KBRgEjJ34YbI_YkRQZV:2050350839) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/af7dcfc28b2eb2e08a3de98680a803440ea5b833/docs/vaadin-combo-box-basic.adoc#L66)):*
The "vaadin-combo-box" tag name is, well, a name, hence it can't have article, but as it's a lower-case name, can't use it in the beginning of sentence either. Hence the convention is to always write: "The <name> element", or in this case, "The [vaadinelement]#vaadin-combo-box# element..."

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/172)*

magi42:

---

*[docs/vaadin-combo-box-basic.adoc, line 66 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/172#-KBRgEjI1EsbedV7ZF-t:-KBRhpHMm2Oyv7KPe55h:-1133888479) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/af7dcfc28b2eb2e08a3de98680a803440ea5b833/docs/vaadin-combo-box-basic.adoc#L66)):*
Oh Reviewable ate my formatting, it was supposed to read: "`The <name> element`"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/172)*

jounik:
> For the "selection value", to me it seems misleading. There are two properties, "value" and "selectedItem", but only the value fires an event when it changes. Not sure if we should mix them in the documentation.

Actually that was wrong. There are two events fired for this. @jojule, can you elaborate what you think was misleading in the original description about "value changes", and why it would need to be more specifically "selection value changes"?

jounik:
Fixed the latter change, I rephrased it in 5145bff8c461e5aeda9318008ba1e59719065367

For the former change (about "selection value change"), any unclear issues regarding it should be handled in https://github.com/vaadin/components-team-tasks/issues/91

PR #174, opened 2016-02-27
Fix infinite loop. Fixes #173

### LINKED ISSUE #173: Infinite loop when a shared bound value is set to null
This code fails when 'name' is undefined, see attached screenshot

```
 <vaadin-combo-box items="[[robots]]" value="{{name}}" label="Select a Robot"></vaadin-combo-box>

```
screen shot 2016-02-27 at 22 51 17

manolo:
[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/174)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -581,7 +581,8 @@ enable usage within an `iron-form`.
 
     _valueChanged: function(value) {
       if (value === undefined || value === null) {
-        this.value = '';
+        // Use 'set' instead of assign to avoid loop.
+        this.set('value', '');
       }
 
       var valueIndex = this._indexOfValue(value);


manolo:
Seems that it happens in non common cases, when data binding is composed by multiple actors, this is my failing example:
https://gist.github.com/manolo/aa8afdbfc0a63d5a6982

Saulis:
Thanks, I can see the problem now. Unfortunately this PR doesn't seem to fix the issue, will investigate more.

Saulis:
I'll close this PR as it looks like a fix needs to be filed to paper-input

manolo:
I need this fix for my demo, otherwise it causes a loop in chrome and safari goes absolutly frozen and even the computer.
In my demo I had to return undefined which is Ok but breaks tests.
I don't thing we need to set value to '' but probably it could be discussed

manolo:
Closing since this fix is not ideal.

PR #175, opened 2016-02-28
Use scroll height to limit overlay height

### LINKED ISSUE #169: Dropdown height limited by body height
As `<vaadin-combo-box>` calculates the dropdown height and position with document's `body`, small body height shows a very small dropdown.
combo-box-demo

Saulis:
Fixes #169 

Originally, plain window.innerHeight is not enough to determine the max-height
for the overlay, because Safari fires scroll events when overscrolling the
body which in turn makes window.innerHeight grow.

Instead, we need a hard limit at the bottom. Body scrollHeight doesnt change (that much)
when scrolling so it works nicely.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/175)

### DIFF
--- a/vaadin-overlay-behavior.html
+++ b/vaadin-overlay-behavior.html
@@ -69,7 +69,7 @@
     _maxHeight: function(targetRect) {
       var margin = 8;
       var minHeight = 116; // Height of two items in combo-box
-      var bottom = Math.min(window.innerHeight, document.body.getBoundingClientRect().bottom);
+      var bottom = Math.min(window.innerHeight, document.body.scrollHeight - document.body.scrollTop);
 
       if (this._alignedAbove) {
         return Math.max(targetRect.top - margin, minHeight) + 'px';


jounik:
Nice catch – didn’t realize body.scrollHeight equals the viewport height even thought the body’s offsetHeight is lower.

platosha:
This works, but feels like we should also prevent the overlay from popping above and resize over the document top when overscrolling above the top. Please consider adding something like this:

```patch
diff --git a/vaadin-overlay-behavior.html b/vaadin-overlay-behavior.html
index 4e18b95..e527c5d 100644
--- a/vaadin-overlay-behavior.html
+++ b/vaadin-overlay-behavior.html
@@ -69,10 +69,11 @@
     _maxHeight: function(targetRect) {
       var margin = 8;
       var minHeight = 116; // Height of two items in combo-box
-      var bottom = Math.min(window.innerHeight, document.body.scrollHeight - document.body.scrollTop);
+      var bottom = Math.min(window.innerHeight,
+          document.body.scrollHeight - Math.max(document.body.scrollTop, 0));
 
       if (this._alignedAbove) {
-        return Math.max(targetRect.top - margin, minHeight) + 'px';
+        return Math.max(targetRect.top - margin + Math.min(document.body.scrollTop, 0), minHeight) + 'px';
       } else {
         return Math.max(bottom - targetRect.bottom - margin, minHeight) + 'px';
       }
@@ -102,7 +103,11 @@
     },
 
     _shouldAlignAbove: function() {
-      var spaceBelow = (window.innerHeight - this.positionTarget.getBoundingClientRect().bottom) / window.innerHeight;
+      var spaceBelow = (
+          window.innerHeight -
+          this.positionTarget.getBoundingClientRect().bottom -
+          Math.min(document.body.scrollTop, 0)
+        ) / window.innerHeight;
 
       return spaceBelow < 0.30;
     }
```

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/175#-:-KBmFAp4wI_zTIxro_uM:734591540)*

jounik:
Verified on OS X Safari and iOS Safari.

jounik:
Crap, didn’t notice the comment from @platosha, sorry about that! :frowning:

PR #195, opened 2016-03-21
Pass through for always float label boolean for paper-container

jermsv1:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/195)


Saulis:
Hi there, and thanks for your interest for the component!

Regarding the changes in the PR:

The `alwaysFloatLabel` property is something that is tightly coupled with `paper-input` which is the currently used input element we have in place. Exposing the property from combo-box would couple it to `paper-input` even more and we want to keep them as decoupled as possible. 

Instead, to provide customisability for the users, our plans are to provide means for the users to inject their own custom input element, please see #198 for details and comment your thoughts about it.

For the `focus` events you've included: combo-box does currently already fire `opened-changed`, `vaadin-dropdown-closed` and `vaadin-dropdown-opened` events, which I think should suffice in most use cases. If they are not enough, I would kindly ask you to file an issue where we could discuss the potential use cases you might have. Thank you!

I'm going to close the PR for now, but please feel free to reopen or comment if you feel that I missed something important.

PR #200, opened 2016-03-25
Possible fix for #106 (filter highlighting for items)

### LINKED ISSUE #106: Highlight matches in the filtered list
Add a means to control HOW the filtered list is rendered.

ie: in our app, we want to highlight portions of the matched items as to better show WHERE the matches occur in each item.

ktiedt:
Fixes #106

Add support for matched/unmatched CSS in overlay items

Modify the iron-list template to break item labels into spans based on filter text and style them appropriately.

---

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/200)

### DIFF
--- a/vaadin-combo-box-overlay.html
+++ b/vaadin-combo-box-overlay.html
@@ -53,6 +53,27 @@
     #selector .item[selected] {
       color: var(--primary-color);
     }
+
+    .item {
+      /* prevent white space between spans */
+      font-size: 0px;
+    }
+
+    .matched-piece {
+      font-size: 13px;
+      text-overflow: ellipsis;
+      /* Prevents stripping of whitespace. */
+      white-space: pre;
+      @apply(--vaadin-combo-box-overlay-matched-text);
+    }
+
+    .unmatched-piece {
+      font-size: 13px;
+      font-weight: bold;
+      text-overflow: ellipsis;
+      white-space: pre;
+      @apply(--vaadin-combo-box-overlay-unmatched-text);
+    }
   </style>
 
   <template>
@@ -73,7 +94,11 @@
               role$="[[_getAriaRole(index)]]"
               aria-selected$="[[_getAriaSelected(_focusedIndex,index)]]"
               focused$="[[_isItemFocused(_focusedIndex,index)]]">
-              [[getItemLabel(item)]]
+              <template is="dom-repeat"
+                items="[[_computePieces(item, _filter)]]"
+                as="piece">
+                <span class$="{{piece.styleClass}}">{{piece.content}}</span>
+              </template>
           </div>
         </template>
       </iron-list>
@@ -114,6 +139,11 @@
 
       _items: {
         type: Object
+      },
+
+      _filter: {
+        type: String,
+        value: ''
       },
 
       _focusedIndex: {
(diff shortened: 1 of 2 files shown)


manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-combo-box-overlay.html, line 72 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KE7wZUKU1U9djjqm6eY:-KE7wZUKU1U9djjqm6eZ:1122831141) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L72)):*
Why unmatched is bolded, I would expect matched part to be emphasised 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-combo-box-overlay.html, line 217 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KE7y6R8IrydKkpMEeeq:-KE7y6R8IrydKkpMEeer:28630143) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L217)):*
My first feeling is that this is too much code to maintain here for a small feature, I would rather use a regular expression like:

```
_computePieces: function(text, filter) {
   return text.replace(RegExp(filter, 'gi'), '<strong>$&</strong>')
}
```
styles would be simpler because you only have to consider the matched part. 
 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

manolo:
It would be nice to have a test

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-:-KE7z_8URxOeWxC5m7Dv:1610711813)*

ktiedt:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-combo-box-overlay.html, line 72 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KE7wZUKU1U9djjqm6eY:-KE84OCsy2h7DveJd_Sb:2116033935) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L72)):*
This matches the pattern I have been working with, the user *knows* what they typed, so bold draws attention to the remaining parts of the unmatched text. Ultimately I don't care what the default styles are, as long as they are pluggable with variables.

---

*[vaadin-combo-box-overlay.html, line 217 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KE7y6R8IrydKkpMEeeq:-KE85-c8tuAm8v5asDLz:-1783700559) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L217)):*
Yes this is possibly a different way to achieve this. We needed clear control over both styles however which made the approach I took make sense. It could quite possibly be paired down to smaller code as is, I leveraged some existing work we had done towards this because we needed a quick solution.

Ultimately there is no harm in iterating and improving the code at a later time. Honestly I expected that to happen if this was merged in.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

Saulis:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-combo-box-overlay.html, line 217 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KE7y6R8IrydKkpMEeeq:-KE85yNFDEhUVpUvjfQ8:1388196670) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L217)):*
Yeah I agree with @manolo here, we should be using Regexp here. I would even consider wrapping it in a reusable custom element that you easily give the filter and text values to and use style mixins to style the matched/unmatched parts.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

platosha:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-combo-box-overlay.html, line 59 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KEAbD7H19J4N6yYKzao:-KEAbD7H19J4N6yYKzap:264857831) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L59)):*
Would be better to remove the white space inside `dom-repeat` instead of tricking with font-size.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

manolo:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-combo-box-overlay.html, line 59 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200#-KEAbD7H19J4N6yYKzao:-KEAfUapWXFtKTzQ1Vbq:-1898123050) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/37a449354debf630a7fc515d48d633e700e10dfa/vaadin-combo-box-overlay.html#L59)):*
Actually, it should be better remove the dom-repeat and just return the string, it's should be faster a simple string.join vs having to instanciate a template, etc

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/200)*

vaadin-combo-box-overlay.html
```diff
@@ -184,6 +214,62 @@
...
+      // If the filter is empty, return a single piece that contains the entire text.
+      if (!filter) {
+        // If no filter, use the matched CSS so not all items show up as unmatched.
+        return [this._createPiece(text, true)];
+      }
+
+      // This array stores wherever the text should be highlighted.
+      var highlightArray = [];
+      var startIndicesOffilterInText = this._getStartIndices(filter, text);
+      startIndicesOffilterInText.forEach((startIndex) => {
```
jounik: Should use ES5 for all code. This doesn’t work in Safari at least (`=>`).

jounik:
Thanks for the PR!

Some things to consider for this feature, that we might affect how we want to implement this:
- It should work with custom item templates, meaning any custom HTML. With this in mind, I think it’s more feasible to just mark the matching occurrences rather both matching and unmatching strings.
- Custom filtering, if something else than "contains" is used

Everything else about this feature is internal implementation details, except the CSS mixins we expose and maybe something related to custom filtering, so those need to be considered a bit more carefully before we accept this patch.

manolo:
Since we already have item templates, probably this feature could be achieved using them.
@Saulis is this PR still relevant?

Saulis:
@manolo yes, I think this PR should be closed for now. I've also posted a workaround in #106 how to achieve highlighting with using templates.

PR #214, opened 2016-04-12
Don't blur the input field when closing dropdown on touch devices

### LINKED ISSUE #203: combo-box loses focus when value selected (Win10/Chrome)
combo_enter

platosha:
Fixes #203

This change removes automatic input blur when closing the dropdown on touch devices, keeping blur on touch start in the overlay. 

- Tapping item in the overlay behaviour: the same as in master. This still blurs the input and hides the keyboard.
- Enter key behaviour: different. Previously it blurred the input on touch devices, now just selects the focused item and closes the overlay.
- Clicking the item with mouse: works the same as in master, blurs the input.

Since the enter key behaviour is different, it could be an UX regression. Did lightweight UX test with Denis on iPhone, findings:
- The most common way to select items on touch device is tapping the item, not the enter key. Even when virtual keyboard is inovolved in filtering the items.
- It's not required for the enter key to close the keyboard on touch device.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/214)

### DIFF
--- a/test/toggling-dropdown.html
+++ b/test/toggling-dropdown.html
@@ -122,15 +122,6 @@
 
            expect(focusedInput).to.be.null;
           });
-
-          it('should blur input on close', function() {
-           combobox.$.input.focus();
-
-           combobox.close();
-
-           var focusedInput = Polymer.dom(combobox.root).querySelector('input:focus');
-           expect(focusedInput).to.be.null;
-          });
         } else if (!safari) { // TODO: Sauce Labs bug doesn't allow Safari in OS X to focus.
           it('should focus input on dropdown open', function() {
             var focusedInput = Polymer.dom(combobox.root).querySelector('input:focus');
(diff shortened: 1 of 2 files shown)


jounik:
The "Clicking the item with mouse: works the same as in master, blurs the input." feels wrong on desktop/mouse. For touch devices it makes sense, as it closes the keyboard at the same time. But for desktop/mouse, I would expect the focus to stay in the combo box.

Not saying we need to change the behaviour now, just thinking it might be something we need to revisit later if users start seeing it as an issue.

tomivirkki:
:lgtm: 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/214#-:-KF9FuXpaua8IQwTOy8W:980077423)*

manolo:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/214#-:-KF9UYgnVbKX8OeGy12M:1417088721)*

PR #229, opened 2016-04-25
Allow the developer to replace the button icons

### LINKED ISSUE #196: Allow developers to change the toggle button icon
As a workaround for situations like #191

Consider the use case of hiding the toggle button completely at the same time.

jouni:
Fixes #196

Remove the dependency to iron-icon(s), to make it easy to replace the
icons with CSS mixins by overriding the background-image, instead of
changing the attribute of the iron-icon element.

Revert the toggle button icon back to the dropdown arrow at the same
time, as the “menu” icon was a failed experiment.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/229)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,8 +22,6 @@
     "iron-flex-layout": "PolymerElements/iron-flex-layout#^1.0.3",
     "iron-form": "polymerelements/iron-form#^1.0.10",
     "iron-form-element-behavior": "polymerelements/iron-form-element-behavior#^1.0.4",
-    "iron-icon": "PolymerElements/iron-icon#^1.0.4",
-    "iron-icons": "PolymerElements/iron-icons#^1.0.3",
     "iron-input": "polymerelements/iron-input#^1.0.5",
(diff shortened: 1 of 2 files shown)


tehapo:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/229#-:-KGG1SvIVbwvMFDme9sN:1417088721)*

jouni:
Don’t merge just yet. I wanted to put this here early, but I’ll still have a quick chat about it with other designers. 

The screenshot needs to be updated also if this change is accepted.

jouni:
Anyone against removing the iron-icons dependency? Should we do the same for vaadin-date-picker?

jouni:
On second thought, this change should be split into two. First change the icon back to the original triangle, then introduce the customizability (and remove the dependency to iron-icons at that point if it seems like a good solution).

Issue #198, opened 2016-03-23
Let the developer customize the input area


jouni:
In order to avoid duplicating a lot of API from paper-input for example, we should allow the developer to wrap any input element with vaadin-combo-box.

Examples:
```html
<vaadin-combo-box items='["jouni", "tomi", "manolo", "tuomas"]'>
  <paper-input label="username">
    <iron-icon icon="mail" prefix></iron-icon>
    <div suffix>@vaadin.com</div>
  </paper-input>
</vaadin-combo-box>
```

```html
<vaadin-combo-box items='["foo", "bar", "baz"]' allow-custom-value>
  <input minlength="3" maxlength="3">
</vaadin-combo-box>
```

jouni:
Instead of wrapping, it could also work by reference, using an attribute (e.g. `for-input`):
```html
<paper-input label="username" id="username">
  <iron-icon icon="mail" prefix></iron-icon>
  <div suffix>@vaadin.com</div>
</paper-input>
<vaadin-combo-box for-input="#username" items='["jouni", "tomi", "manolo", "tuomas"]'></vaadin-combo-box>
```

jouni:
##### API draft (just my 5 cents)

This tries to consider the upcoming item template (#165) and multi-select (#88) features as well, how those might affect the API. I might be overcomplicating things needlessly at this point, but I just can’t help it, I need to convince myself that we can are not implementing anything that would limit our options in the future 😄 

The following code should result in something like the following mockup: 
![combo-box-mockup](https://cloud.githubusercontent.com/assets/66382/15705159/b49cb16a-27f6-11e6-8546-19751fb21f17.png)

```html
<vaadin-combo-box allow-multi-select allow-custom-input>

  <!-- Notice the class="input" -->
  <paper-input label="Recipients" class="input">
    <iron-icon icon="email" prefix></iron-icon>
    <!-- Notice the class="selected-items" (selected Items/chips/tokens will be rendered here) -->
    <div class="selected-items" prefix></div>
    <!-- Notice the class="clear-button" -->
    <paper-icon-button icon="clear" suffix class="clear-button"></paper-icon-button>
  </paper-input>

  <!-- Notice the class="overlay-item" -->
  <template class="overlay-item">
    <paper-icon-item>
      <div class="avatar blue" item-icon></div>
      <paper-item-body two-line>
        <div>[[item.name]]</div>
        <div secondary>[[item.email]]</div>
      </paper-item-body>
[...]

PR #232, opened 2016-04-29
Fire event when user types in input

pbochis:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/232)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -501,6 +501,7 @@ Custom property | Description | Default
      */
 
     _inputValueChanged: function(e) {
+      this.fire('input-value-changed', this.$.input.bindValue);
       if (this._filter === this.$.input.bindValue) {
         // Filter and input value might get out of sync, while keyboard navigating for example.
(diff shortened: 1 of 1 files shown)


Saulis:
Hi and thanks for the PR!

We've planned on adding support for custom input elements (see #198) which would allow devs to expose any events they need. Exposing `input-value-changed` event would conflict with that approach, so we look into that once we start implementing the support.

Meanwhile, you can get the same behaviour currently at least with two approaches: manually adding an `valueChanged` listener to `combobox.$.input` or adding a `keyup` listener to `<vaadin-combo-box>` and fetching the `value` using `event.target.value` for example. 

If this doesn't sound good to you, please elaborate on your use case and the reasons why not. Thanks!

platosha:
Hi!

Now that we’ve implemented the light DOM input element support with the `vaadin-combo-box-light` element, you should be able to bind your events this way. This feature is going to be released in 1.1 version, but you can try it in the master branch already.

I’m closing this PR. Please file an issue if your needs are not fulfilled.

PR #240, opened 2016-05-12
Always hide clear icon

JMuratha:
I added this functionality because i didn't want to allow users to delete their input/selection (needed it for required fields)

Maybe you can also include it in the official version.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/240)

### DIFF
--- a/demo/basic.html
+++ b/demo/basic.html
@@ -45,7 +45,7 @@ layout: page
     </p>
 
     <code-example source>
-      <vaadin-combo-box demo label="Element"></vaadin-combo-box>
+      <vaadin-combo-box demo label="Element" required></vaadin-combo-box>
 
       <code demo-var="combobox">
         HTMLImports.whenReady(function() {
--- a/test/selecting-items.html
+++ b/test/selecting-items.html
@@ -12,151 +12,151 @@
 
 <body>
 
-<test-fixture id="combobox">
-  <template>
-    <vaadin-combo-box></vaadin-combo-box>
-  </template>
-</test-fixture>
+  <test-fixture id="combobox">
+    <template>
+      <vaadin-combo-box></vaadin-combo-box>
+    </template>
+  </test-fixture>
 
 
-<script>
-  describe('selecting a value', function() {
-    var combobox;
-    var input;
-    var selector;
-    var valueChangedSpy;
+  <script>
+    describe('selecting a value', function () {
+      var combobox;
+      var input;
+      var selector;
+      var valueChangedSpy;
 
-    beforeEach(function() {
-      combobox = fixture('combobox');
-      combobox.items = ['foo', 'bar'];
-      input = combobox.$$('input');
-      selector = combobox.$.overlay.$.selector;
-      valueChangedSpy = sinon.spy();
-      combobox.addEventListener('value-changed', valueChangedSpy);
-    });
-
-    it('should select by using the selector', function() {
-      selector.selectedItem = 'foo';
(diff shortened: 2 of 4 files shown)


Saulis:
Hi!

 If you need to prevent users from submitting empty values, I'd suggest using the `required` attribute insidfe an `iron-form`.

See the example "Submitting combo-box value through a form" at https://vaadin.com/elements/-/element/vaadin-combo-box

You see, hiding the clear icon isn't enough, input/selected value can still be cleared by using the keyboard.

JMuratha:
I know how to prevent submiting to form without a value, but the clear icon suggest the user can have an empty value so i wanted to hide it.

But i didn't knew that by deleting the value by hand no item gets selected.
I must somehow missed that when i was testing the `vaadin-combo-box`.

I will implemented a new property then to force a value to be selected.

If you want i can create a pull request for this feature too.

Saulis:
OK I see.

To get the ball rolling, I have some proposals on how we could approach this use case.

##### 1. Add a feature to enable the user to give a custom element for the clear icon
Pros:
  - Enables full customisation of the clear icon with minor changes to the component

Cons:
  - Dev needs to manually bind the `tap` event to something like `combobox.clear()`
  - Dev needs to manually handle `value-changed` events when the user has cleared the input field with their keyboard.

##### 2. Protect value from clearing always when `required` is `true`.
Pros:
  - No API changes at all
  - Easy to implement
  - New behaviour is easy to communicate to devs

Cons:
  - UX might be poor for users expecting to be able to clear a field *before submit* even if it is `required`
  - This behaviour would be forced upon all users

##### 3. Add a new property `allow-clear`or `prevent-clear` or similar
( this is in essence what @JMuratha is proposing, expect for the property names and the coupling between `required` and `hideClearIcon` properties)
Pros: 
  - Behaviour should be easy to communicate to devs
  - Can be used independently from `required` and `iron-form` 

Cons:
  - Adds yet another configurable property to our API.

Comments, suggestions? @vaadin/vaadin-elements

JMuratha:
I already made an additional change to my fork which is bound to the `required` attribute.

- When required is true the first item gets automaticaly selected.
- The clear button now "resets" the input (selecting the first item).
- When the user is deleting the input (with backspace) and focuses another element, the first item gets selected.

Now i only need to change the behavior of the clear button, so when pressing the clear button you will have the empty input focused, which will be no big deal. 

Seems like the best implementation for my use case.

This implementation is exactly as your 2. proposal.

jouni:
@JMuratha, we’ve started designing and prototyping #198 more closely, so now would be a great time to start giving more feedback how it’s going to look like and if it would satisfy your use case. Check out the comments in that issue and let’s continue the discussion there. Thanks!

JMuratha:
@jouni 

(im continuing here because i don't think it would fit the other discussion)

#198 doesn't actually seem to let me customize the `vaadin-combo-box` the way i want it to behave.

Since the last time i wrote here i made several changes to my fork.
Now it behaves like this when the required property is set:

- when initialising the `vaadin-combo-box` without a value, the first item gets selected automatically
- when initialising the `vaadin-combo-box` with a value, this value gets selected if available, else the first value gets selected
- when pressing the clear icon, the input gets cleared and focused
- when leaving/bluring the `vaadin-combo-box` with an invalid/empty value, the first item gets selected

If you also want to implement this behavior the best way propably would be  to add an additonal attribute like `force-value-selection` or something like this.

PR #244, opened 2016-05-13
Allow 'clear' to be disabled by adding a 'disable-clear' attribute.

depeele:
Simply adding a `disable-clear` attribute to the `<vaadin-combo-box>` component will disable the clear button (and functionally the clear functionality).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/244)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -293,6 +293,16 @@ Custom property | Description | Default
       itemValuePath: {
         type: String,
         value: 'value'
+      },
+
+      /**
+       * Disable the 'clear' button.
+       *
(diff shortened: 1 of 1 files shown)


Saulis:
Related with #240

jouni:
Also related with #198

jouni:
We’ve started designing and prototyping #198 more closely, so now would be a great time to start giving more feedback how it’s going to look like and if it would satisfy this use case as well. Check out the comments in that issue and let’s continue the discussion there. Thanks!

Saulis:
Closing this as input area can be customized since `v1.1.0`

Issue #165, opened 2016-02-22
Custom item template support


jouni:
Split from #96, where this was already researched.
### Research report

Exposing items template is one step beyond the scope of this task. Despite that, it was also prototyped to get the idea of its complexity level.

The prototype branch contains a draft but working implementation: https://github.com/vaadin/vaadin-combo-box/tree/proto/custom-item-template

Exposing an item template is a complex task as well (roughly at the sample level, I’d say). While possible per se, still it requires to introduce a separate `vaadin-combo-box-item` element and expose the `iron-list` in the light DOM of the overlay. Deep explanation of the problems below.
- To achieve that, the first idea was to set the entire `iron-list` item template with the template provided in the combo-box light DOM.
  - Turned out to be impractical: we still need to have an `.item` node defined inside the overlay, assigning model and state attributes (like `id` with index, `selected`, `focused`, ARIA `role` and `aria-selected`) and basic styles to it.
- The second idea was: instead of replacing the entire item template, try to compose the built-in item template with the one from the outside, inserting it manually inside the `.item` wrapper. 
  - Hard to implement. The puzzle is: when registering the overlay element, Polymer parses the contents of `<template>` nodes as well, extracting bindings into the annotations. After this parsing at the register time, it’s impossible to modify the contents of the item template.
    - Tried to disable parsing bindings on the overlay register by adding `<template preserve-content>` (undocumented Polymer feature) attribute to the `iron-list` item template. This makes it possible to modify the item template, but at the same time effectively prevents the `.item` bindings to the overlay methods from being parsed and working. Hence the `focused` indication and ARIA attributes don’t work.
- The working idea: instead of manually modifying the item template, change the `div.item` into a separate `vaadin-combo-box-item` element, that would support custom template in the `template` property. This is currently implemented in the prototype branch.
  - For the item element, we need to `templatize()` it with either a given template or a default item template, stamp a template instance and append it to the light DOM. The changes to the bound `item` model should be updated on the stamped template instance manually.
  - `iron-list` exposes the item nodes in the light DOM. But still the overlay also needs to expose the `iron-list` in the light DOM, in order to be targetable with global CSS.
### Related issues
- #106

web-padawan:
@jounki is this task still in the backlog?
Actually I have to use ugly code with deprecated selectors like this to apply some styling to items:

```css
            body > #overlay::shadow .item {
                white-space: normal;
            }

            body > #overlay::shadow .item[selected] {
[...]

PR #246, opened 2016-05-13
Add a --vaadin-combo-box-overlay-item css mixin

depeele:
Add a --vaadin-combo-box-overlay-item css mixin to allow simple theme-based styling of combo-box items.

Example:

``` css
  vaadin-combo-box {
    --vaadin-combo-box-overlay-item: {
      padding: 0.5em;
    }
  }
```

---

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/246)

### DIFF
--- a/vaadin-combo-box-overlay.html
+++ b/vaadin-combo-box-overlay.html
@@ -50,6 +50,8 @@
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;
+
+      @apply(--vaadin-combo-box-overlay-item);
     }
 
     #selector:not([touch-device]) .item:hover,


Saulis:
Related with #165

jouni:
This looks okay to me, if #165 will work so that the custom template will be placed inside the current `.item` element. But if the custom template will completely replace the `.item` element, then I’m not sure if this is a necessary addition, as you could then provide a custom template and style that instead.

web-padawan:
Although implementing custom item template is a good idea, there should also be a number of available mixins for styling toggle and close buttons, and the input container itself. 
So I'll create a separate PR for that.

Saulis:
Keeping this still on ice until we get item templates forward since it will most likely break this mixin.

manolo:
@Saulis items templates already released, is this PR still relevant?

jouni:
Since this PR is already over a year old, I’m closing it. Please add comments if you think we need to take another look.

PR #251, opened 2016-06-06
Fix the notifyPath call to have the explicit `undefined` value argument

### LINKED ISSUE #250: Master tests are broken
Although in local tests pass, in Travis this test fails : 
test/vaadin-combo-box-properties.html » properties » items property » should update items list on mutation

It may be a flaky test?

platosha:
Fixes #250

The second and the third arguments in Polymer.Base.notifyPath are now optional since Polymer v1.5.0. Because of that, the behavior when called with a missing argument has changed. We have to use explicit `undefined` value argument from now on.

More about the change: https://github.com/Polymer/polymer/pull/3473

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/251)

### DIFF
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -580,7 +580,7 @@ Custom property | Description | Default
       //TODO: There is something wrong with observers and using set function
       // see here: https://github.com/Polymer/polymer/issues/3254
       //As a workaround setting/notifying the observer with the value undefined helps
-      this.$.overlay.notifyPath('_items');
+      this.$.overlay.notifyPath('_items', undefined);
       this.$.overlay.set('_items', items);
 
       var valueIndex = this._indexOfValue(this.value, items);


tehapo:
:lgtm:

> #### Fix the notifyPath call to have the explicit \`undefined\` value argument
> 
> Fixes #250
> 
> 
> 
> The second and the third arguments in Polymer.Base.notifyPath are now optional since Polymer v1.5.0. Because of that, the behavior when called with a missing argument has changed. We have to use explicit `undefined` value argument from now on.
> 
> 
> 
> More about the change: https://github.com/Polymer/polymer/pull/3473
> 
> 
> 
</details>


---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/251#-:-KJZvlmBZ_7SW-ufvX4G:1417088721)*

PR #252, opened 2016-06-06
Fix/239

### LINKED ISSUE #239: Input disabled on Edge if combo-box has selected value
Tested on Edge with a Surface 4 tablet. If a combo box has a pre selected value when focusing it, the cursor gets inserted into the field, but keyboard does not show. Once the combo-box is cleared, you are able to input text.

https://drive.google.com/open?id=0B-8uh2URmjlvMFhJOXNpdjBCYms

platosha:
Fixes #239

The change switches the opening of the combobox to be asynchronous when the user taps on the input field or on the label.

Makes MS Edge on Surface to always show the virtual keyboard.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/252)

### DIFF
--- a/test/toggling-dropdown.html
+++ b/test/toggling-dropdown.html
@@ -36,19 +36,27 @@
     });
 
     describe('opening', function() {
-      it('should open by tapping label', function() {
+      it('should open asynchronously by tapping label', function(done) {
         Polymer.Base.fire('tap', {}, {
           bubbles: true,
           node: combobox.$$('paper-input-container label')
         });
 
-        expect(combobox.opened).to.be.true;
+        expect(combobox.opened).to.be.false;
+        Polymer.Base.async(function() {
+          expect(combobox.opened).to.be.true;
+          done();
+        }, 1);
       });
 
-      it('should open by clicking input', function() {
+      it('should open asynchronously by clicking input', function(done) {
         combobox.$$('paper-input-container input').fire('tap');
 
-        expect(combobox.opened).to.be.true;
+        expect(combobox.opened).to.be.false;
+        Polymer.Base.async(function() {
+          expect(combobox.opened).to.be.true;
+          done();
+        }, 1);
       });
 
       it('should open by clicking icon', function() {
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -145,7 +145,7 @@ Custom property | Description | Default
         invalid="[[invalid]]"
         opened$="[[opened]]"
         hideclear$="[[_hideClearIcon(value, opened)]]">
-      <label id="label" on-tap="open">[[label]]</label>
+      <label id="label" on-tap="_openAsync">[[label]]</label>
       <input
           is="iron-input"
           id="input"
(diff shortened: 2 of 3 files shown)


manolo:
Reviewed 1 of 1 files at r3.
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-dropdown-behavior.html, line 60 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252#-KJeLVzOfMQo_teO9Irt:-KJeLVzOfMQo_teO9Iru:-586430394) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/23cab63c4866c7602edf52e9b6ef5c6a872229b2/vaadin-dropdown-behavior.html#L60)):*
> ```HTML
>      * Opens the dropdown list asynchronously.
>      */
>     openAsync: function() {
> ```

Do we need two public functions? open and openAsync ?  in the API.
Why not make open be async instead?
Actually the user has another public API this.opened = true to execute sync.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252)*

platosha:

---

*[vaadin-dropdown-behavior.html, line 60 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252#-KJeLVzOfMQo_teO9Irt:-KJeNlAxVG0WF5aeepqd:-2048829261) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/23cab63c4866c7602edf52e9b6ef5c6a872229b2/vaadin-dropdown-behavior.html#L60)):*
> Do we need two public functions? open and openAsync ?  in the API.
> Why not make open be async instead?
> Actually the user has another public API this.opened = true to execute sync.
</details>
Good questions.

We don’t need public function for that, changed to private.

Also, this seems to be related with the specific use case in the combo-box. In general, might not be needed for any dropdown. I moved the method to the vaadin-combo-box.html file from the behaviour.

Why not make `open` be async? `open()` used in multiple tests, which rely on the synchronous behaviour. In fact, we spent some time trying to achieve this, in order to simplify these tests and remove the asynchronous intervals. Changing `open()` to be asynchronous again requires changes in approx. 10% of tests. Changing tests could be dangerous. I don't feel like we have a good reason for that.

Additionally, it doesn't seem like there is a bug in `open()` that requires any changes inside. I haven’t found anything incorrect with the method itself. Looks like the issue is an effect of doing too much DOM work synchronously in the event hander. For this reason, `open()` needs to be invoked asynchronously.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252)*

manolo:
:lgtm:

> #### Fix/239
> 
> Fixes #239
> 
> 
> 
> The change switches the opening of the combobox to be asynchronous when the user taps on the input field or on the label.
> 
> 
> 
> Makes MS Edge on Surface to always show the virtual keyboard.
> 
> 
> 
</details>


---

Reviewed 2 of 4 files at r2, 1 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252#-:-KJeSoirvKB4c6ReNe4o:1417088721)*

manolo:
Reviewed 1 of 2 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/252)*

PR #254, opened 2016-06-09
Introduce input customization feature

### LINKED ISSUE #198: Let the developer customize the input area

tehapo:
Fixes #198 

**IMPORTANT! Please don't merge this PR just yet. We need to rebase, clean up and squash the commit history before merging.**

This is quite a big change and therefore I tried to list all the main points here.

Ground work
----
* Our `iron-icon` + `paper-ripple` combo is now replaced with `paper-icon-button`.
* Following attributes are now dynamically added/removed to the host element to enable styles based on the component state.
  * `opened`
  * `has-value`
  * `readonly`
  * `disabled`

Customization feature
----
* Replacing the `paper-icon-buttons` can be done by adding child elements with class names of `toggle-button` and/or `clear-button`.
* Two new class names are exposed for styling.
  * `small` – makes the icon slightly smaller (applied to the default clear icon, only affects a `paper-icon-button` element).
  * `rotate-on-open` – makes the button rotate with a transition when dropdown is opened (applied to the default toggle icon).

Example.
```html
<vaadin-combo-box>
  <paper-icon-button icon="remove-circle" class="clear-button"></paper-icon-button>
  <paper-icon-button icon="search" class="toggle-button"></paper-icon-button>
</vaadin-combo-box>
```

* For replacing the whole input we introduced a new element called `vaadin-combo-box-light`.
  * This split will avoid unnecessary imports when using a customized version.
  * All logic from `vaadin-combo-box` was refactored to new behavior called `vaadin-combo-box-behavior`.
  * The `vaadin-combo-box-light` expects to have a child element matching the selector `input[is="iron-input"],paper-input,.paper-input-input,.input`.
  * The input element must have some two-way data-bindable property for the current value.
  * By default the value property is expected to be `bind-value` (to be compatible with `iron-input`).
  * Value property name can be changed with the `attrForValue` property.

Examples.
```html
<vaadin-combo-box-light>
  <input is="iron-input">
</vaadin-combo-box-light>

<vaadin-combo-box-light attr-for-value="value">
  <paper-input label="Elements" class="input">
    <iron-icon icon="toll" prefix></iron-icon>
    <paper-button suffix class="clear-button">Clear</paper-button>
    <paper-button suffix class="toggle-button">Toggle</paper-button>
  </paper-input>
</vaadin-combo-box-light>
```

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/254)


manolo:
Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[bower.json, line 37 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq8PCYE_w29euosvhT:-KJq8PCZ9_d0sVTEvrdr:925779229) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/bower.json#L37)):*
> ```JSON
>     "paper-styles": "polymerelements/paper-styles#^1.1.2",
>     "iron-a11y-announcer": "polymerelements/iron-a11y-announcer#^1.0.4",
>     "paper-icon-button": "polymerelements/paper-icon-button#^1.1.1"
> ```

so we need latest ?  I think 1.0.x would be enough

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KJq92ywY8QRLGsviQ6e:-100340281) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
> ```HTML
>       /**
>        * An array of values to be displayed as options in the dropdown. The
>        * options could be either `String` or `Object`.
> ```

Should be documented as {(Array<String>|Array<Object>)} , but actually String is a valid object, so I would remove the second sentence

---

*[vaadin-combo-box-behavior.html, line 21 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9RtvHwBooieXIZAe:-KJq9RtvHwBooieXIZAf:-885398741) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L21)):*
> ```HTML
> 
>       /**
>        * If `true`, user can input a value that is not present in the items list.
> ```

'The user can type ...'

---

*[vaadin-combo-box-behavior.html, line 33 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9bJp-Ka0Gl4LSOWy:-KJq9bJp-Ka0Gl4LSOWz:-155593316) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L33)):*
> ```HTML
>       /**
>        * A read-only property indicating whether this combo box has a value
>        * selected or not. Can be used for example in styling of the component.
> ```

It can ...  ... styling the component

---

*[vaadin-combo-box-behavior.html, line 74 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9p4U_XC0uETG0aKL:-KJq9p4U_XC0uETG0aKM:1320784262) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L74)):*
> ```HTML
> 
>       /**
>        * Path for value of the item. If `items` is an array of objects, the
> ```

for the value

---

*[vaadin-combo-box-behavior.html, line 78 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9uTSK1PMh2b0MxVQ:-KJq9uTTcQPWZz0dRpDc:1027669493) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L78)):*
> ```HTML
>        * item.
>        *
>        * The item value is to be used in the `value` property of the combo box,
> ```

'is going to be used' or 'will be used' or 'is used'

---

*[vaadin-combo-box-behavior.html, line 188 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6e_CYx3mza0hLaPt:-KJq6e_CYx3mza0hLaPu:-668255301) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L188)):*
> ```HTML
>     _getItemValue: function(item) {
>       var value = this.get(this.itemValuePath, item);
>       if (value === undefined || value === null) {
> ```

couldn't this be simplified ? `value == null`

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KJq6PCYr8yKqxE0syJ7:-41487978) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
> ```HTML
> 
>     /**
>      * Opening and closing related functionality
> ```

What is this comment about?

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KJqAwzUjqfdn6hfpL0V:1554778362) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
> ```HTML
>      * Reverts back to original value.
>      */
>     cancel: function() {
> ```

Only one public method  in all the behaviour?  I guess it should be private or provide some other, like clear, toggle

---

*[vaadin-combo-box-light.html, line 6 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq67IA3QpVWUQHYt-7:-KJq67IA3QpVWUQHYt-8:497797710) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-light.html#L6)):*
> ```HTML
> <link rel="import" href="vaadin-combo-box-shared-styles.html">
> 
> <dom-module id="vaadin-combo-box-light">
> ```

Please document this element

---

*[vaadin-combo-box.html, line 184 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqEZw4pvomrxbtM4uN:-KJqEZw4pvomrxbtM4uO:-722965528) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L184)):*
> ```HTML
> 
>     attached: function() {
>       this.inputElement = this.$.input;
> ```

why is inputElement var public?

---

*[vaadin-combo-box.html, line 208 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq5NLHMDw7mJbDeLhr:-KJq5NLHMDw7mJbDeLhs:1140617877) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L208)):*
> ```HTML
>     },
> 
>     _preventDefault: function(e) {
> ```

duplicated method here and in the -light, move it to the behavior

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[bower.json, line 37 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq8PCYE_w29euosvhT:-KJsqDKC80fvCGtfWXod:1688845437) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/bower.json#L37)):*
> so we need latest ?  I think 1.0.x would be enough
</details>
Good point. Changed to required version to 1.0.5 and upwards after testing which version is the first one that works without any issues.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box.html, line 184 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqEZw4pvomrxbtM4uN:-KJtiR-XUqcAPHyayU2P:-94005019) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L184)):*
> why is inputElement var public?
</details>
This was decided by @platosha, but I can't see any reason for it to be public anymore. Changed to private.

---

*[vaadin-combo-box.html, line 208 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq5NLHMDw7mJbDeLhr:-KJth21Q_zaJVwME0Q6g:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L208)):*
> duplicated method here and in the -light, move it to the behavior
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-light.html, line 6 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq67IA3QpVWUQHYt-7:-KJtzguk2-EJtJRA_uEU:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-light.html#L6)):*
> Please document this element
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

manolo:
Review status: 0 of 20 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-light.html, line 17 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu5iDak-c9MwL48BQn:-KJu5iDbVH1NdFT-GuU0:1459862358) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L17)):*
> ```HTML
> data-bindable property representing the input value. The property name is expected
> to be `bindValue` by default. See the example below for a simplest possible example
> using `<iron-input>` as the input field.
> ```

Well, to be precise,  this is not a proper  `iron-input` but an element `input` extending it.  Probably any other element extending it should work.

---

*[vaadin-combo-box-light.html, line 25 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6Kt_t8UqRPbAEAij:-KJu6Kta2t2WZmFrK0tz:-1611468950) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L25)):*
> ```HTML
> ```
> 
> If you are using other custom input fields or for example `<paper-input>`, you
> ```

use `like` instead of for example, the word example is used too many times in the same paragraph.

---

*[vaadin-combo-box-light.html, line 27 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6T_KkBldSNhhvLMH:-KJu6T_KkBldSNhhvLMI:-2032185607) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L27)):*
> ```HTML
> If you are using other custom input fields or for example `<paper-input>`, you
> need to define the name of value property with `attrForValue` property.
> See the example below for an example on how to create a custom input field
> ```

Remove the `forn an example` part which is redundant, since it is an example already

---

*[vaadin-combo-box-light.html, line 28 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6dkJtWEDpRH5krC9:-KJu6dkJtWEDpRH5krCA:1668612284) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L28)):*
> ```HTML
> need to define the name of value property with `attrForValue` property.
> See the example below for an example on how to create a custom input field
> with a `<paper-input>` and decorating it with a custom `<iron-icon>` and
> ```

based on a `<paper-input>` decorated with ...

---

*[vaadin-combo-box.html, line 10 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu7g0cyF7DpPqIKsVD:-KJu7g0dT9eggeJsBu4D:1369657931) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box.html#L10)):*
> ```HTML
> `<vaadin-combo-box>` is a combo box element combining a dropdown list with an
> input field for filtering the list of items. If you would like to use a custom
> input field, see [`<vaadin-combo-box-light>`](#vaadin-combo-box-light).
> ```

This sentence it's not clear to me. I would rephrase:

If you want to replace the default input field implementation, you should use the .. element instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KJu8FSFEOkikF37oe8H:-958788276) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
> Should be documented as {(Array<String>|Array<Object>)} , but actually String is a valid object, so I would remove the second sentence
</details>
Not sure if I understand what you mean. Made small adjustments to this to make it more clear.

---

*[vaadin-combo-box-behavior.html, line 21 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9RtvHwBooieXIZAe:-KJu8TtdKl_XMYSIsE1Z:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L21)):*
> 'The user can type ...'
</details>
Done.

---

*[vaadin-combo-box-behavior.html, line 33 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9bJp-Ka0Gl4LSOWy:-KJu8WHXwcxaeYY7_oIf:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L33)):*
> It can ...  ... styling the component
</details>
Done.

---

*[vaadin-combo-box-behavior.html, line 74 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9p4U_XC0uETG0aKL:-KJu8Zor3hnjsEafGYcQ:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L74)):*
> for the value
</details>
Done.

---

*[vaadin-combo-box-behavior.html, line 78 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq9uTSK1PMh2b0MxVQ:-KJu8f-x7rhtv7NhgPyZ:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L78)):*
> 'is going to be used' or 'will be used' or 'is used'
</details>
Done.

---

*[vaadin-combo-box-behavior.html, line 188 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6e_CYx3mza0hLaPt:-KJu8gMlhs0v6P8UPJ9e:834239772) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L188)):*
> couldn't this be simplified ? `value == null`
</details>
Yes, that would work the same, but this is just copied from vaadin-combo-box and I guess it's ok as it is.

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KJu8yv3_rTQ78ijDNUr:63689955) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
> What is this comment about?
</details>
Removed the comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-light.html, line 17 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu5iDak-c9MwL48BQn:-KJuAPFfgPUkT9U1jWj-:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L17)):*
> Well, to be precise,  this is not a proper  `iron-input` but an element `input` extending it.  Probably any other element extending it should work.
</details>
Done.

---

*[vaadin-combo-box-light.html, line 25 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6Kt_t8UqRPbAEAij:-KJuAPCjyeQyqAfcpt4b:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L25)):*
> use `like` instead of for example, the word example is used too many times in the same paragraph.
</details>
Done.

---

*[vaadin-combo-box-light.html, line 27 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6T_KkBldSNhhvLMH:-KJuAOSMCWgw3fhVdjte:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L27)):*
> Remove the `forn an example` part which is redundant, since it is an example already
</details>
Done.

---

*[vaadin-combo-box-light.html, line 28 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu6dkJtWEDpRH5krC9:-KJuAOEPhOVwFadzpa6j:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-light.html#L28)):*
> based on a `<paper-input>` decorated with ...
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 0 of 20 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KJuB6sFJ3uBZH8NH_PZ:-242498894) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
> Only one public method  in all the behaviour?  I guess it should be private or provide some other, like clear, toggle
</details>
This method is copied from current version of `vaadin-combo-box` where it is already public, so we cannot remove this to keep the component backwards compatible.

---

*[vaadin-combo-box.html, line 10 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJu7g0cyF7DpPqIKsVD:-KJuB-fFVeht5WdP6pJV:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box.html#L10)):*
> This sentence it's not clear to me. I would rephrase:
> 
> If you want to replace the default input field implementation, you should use the .. element instead.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

manolo:
Review status: 0 of 20 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KJuSSjf7gxzXttcX6SK:-120706294) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> Not sure if I understand what you mean. Made small adjustments to this to make it more clear.
</details>
Seems that my comment was altered because I added some symbols without quoting.
What I meant is that the way to document different types for the same parameter is :  `{(Array<Object>|Array<String>)}` .
But actually a String is an Object, so I think it is enough if you just say: `An array of Objects to be displayed as options in the dropdown` 

---

*[vaadin-combo-box-behavior.html, line 188 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6e_CYx3mza0hLaPt:-KJuU0qAt9EBLU0PU9ab:1938401875) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L188)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> Yes, that would work the same, but this is just copied from vaadin-combo-box and I guess it's ok as it is.
</details>
I do prefer to leave the minimum code in JS. That makes easy to read and optimises the code. I know this is minimum improvement though but every min contribution is good, so you decide.

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KJuTes7O5X2Xo-Zapo5:-179201174) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> This method is copied from current version of `vaadin-combo-box` where it is already public, so we cannot remove this to keep the component backwards compatible.
</details>
I think we should discuss so what should be the public API of this behaviour. I think if _clear is not public cancel makes not sense to be public.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Reviewed 2 of 20 files at r1.
Review status: 0 of 20 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 39 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7ig-Bg2wpLdrRTsNI:-KK7ig-Bg2wpLdrRTsNJ:-2099797869) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L39)):*
> ```HTML
>         readonly: true,
>         reflectToAttribute: true
>       },
> ```

Add a default value here, maybe? Just to make it more explicit and clear.

---

*[vaadin-combo-box-behavior.html, line 86 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7j2sqsNz5mU6wrJ4o:-KK7j2srrqejEmpFpwaF:1333401146) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L86)):*
> ```HTML
>       },
> 
>       inputElement: Object,
> ```

This became public. Feels like it needs an APIdoc comment then.

---

*[vaadin-combo-box-behavior.html, line 188 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6e_CYx3mza0hLaPt:-KK7jTukaT4LiiJPG2vh:-172232383) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L188)):*
> I do prefer to leave the minimum code in JS. That makes easy to read and optimises the code. I know this is minimum improvement though but every min contribution is good, so you decide.
</details>
The problem with ```value == null``` comparison is that doesn’t pass the linter.

I suggested this once to Sami already. Would prefer the shorter variant, though the current one works for me too.

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KK7kQKfTdgHnjFetVRN:-384377348) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
> I think we should discuss so what should be the public API of this behaviour. I think if _clear is not public cancel makes not sense to be public.
</details>
The method was public in the `vaadin-combo-box` API already. It’s not specific for a particular input element, so it should be in `vaadin-combo-box-light` as well.

Though we do export the behavior, you can’t call its methods directly. It doesn’t work like this. We should rather consider the behavior methods / props / events as the elements’s API, not the behavior's API.

This way it makes a perfect sense to extract the identical combo-box methods to the combo-box-behavior, both public and private ones.

---

*[vaadin-combo-box.html, line 184 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqEZw4pvomrxbtM4uN:-KK7zXmqxkOX-zsmaREv:1221741592) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L184)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> This was decided by @platosha, but I can't see any reason for it to be public anymore. Changed to private.
</details>
It was public to be aligned with `PaperInputBehavior`, which also has a public inputElement property.

Not that the combo-box now has `PaperInputBehavior`, (not the combo-box-light). I thought that it would make sense to override the `inputElement` property coming from the behavior with the correct input for our case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Reviewed 3 of 20 files at r1.
Review status: 1 of 20 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-shared-styles.html, line 73 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK80zRdqnicVIfn1DAQ:-KK80zRdqnicVIfn1DAR:-1001471526) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-shared-styles.html#L73)):*
> ```HTML
>       :host([opened]) paper-input-container paper-icon-button:hover {
>         color: rgba(0, 0, 0, .86);
>       }
> ```

Looks like all the styles above that are scoped to `paper-input-container` wouldn't apply for the combo-box-light. They are limited to the paper-input-container in the shadow DOM of the element, which is never true in case of the light combo-box.

---

*[vaadin-combo-box.html, line 184 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqEZw4pvomrxbtM4uN:-KK82dLYR-pvJKcCmgWx:1448886558) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L184)):*
> It was public to be aligned with `PaperInputBehavior`, which also has a public inputElement property.
> 
> Not that the combo-box now has `PaperInputBehavior`, (not the combo-box-light). I thought that it would make sense to override the `inputElement` property coming from the behavior with the correct input for our case.
</details>
I mean, “*Note* that the combo-box...”

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Reviewed 1 of 20 files at r1.
Review status: 1 of 20 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Reviewed 14 of 20 files at r1, 1 of 1 files at r2, 11 of 11 files at r3, 3 of 3 files at r4, 3 of 3 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KK85KkBKaMtkh1uD33E:1269811229) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
> Seems that my comment was altered because I added some symbols without quoting.
> What I meant is that the way to document different types for the same parameter is :  `{(Array<Object>|Array<String>)}` .
> But actually a String is an Object, so I think it is enough if you just say: `An array of Objects to be displayed as options in the dropdown` 
</details>
Disagree. The `String` case is different from the `Object` here. Strings are valid items as they are, they are not treated as like any objects. Strings should be mentioned explicitly here.

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KK8GT6X08PjJBWq9e8h:-947708558) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> Removed the comment.
</details>
It's a group header for the next few methods. Please bring it back.

---

*[vaadin-combo-box-behavior.html, line 346 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8F9TPUMtQZ6Kt8Clo:-KK8F9TQM6ywdUJWM1gG:854475368) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/dfdf04224bce7cbcd81f0e8a4e09087fcdca7922/vaadin-combo-box-behavior.html#L346)):*
> ```HTML
>     _focusedInput: function() {
>       // Use either light DOM input or one from local DOM.
>       return Polymer.dom(this).querySelector('input:focus') || Polymer.dom(this.root).querySelector('input:focus');
> ```

IMO, this is worse than having two separate methods. Not a blocker though.

---

*[vaadin-combo-box-shared-styles.html, line 42 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK87rVBNWyW2jfjq47E:-KK87rVBNWyW2jfjq47F:832685315) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-shared-styles.html#L42)):*
> ```HTML
>       :host ::content paper-icon-button.clear-button,
>       :host ::content paper-icon-button.toggle-button {
>         position: absolute;
> ```

These absolute positioning styles for the clear and the toggle buttons require complementary equal padding for the input element. This condition is fulfilled for the regular combo-box, but not for the light one.

For example, if someone use `<paper-input-button class="toggle-button">` with the light combo-box, they would have a potential overlap of the icon and the input value. To resolve this, they have to figure out and extract the correct solution from the combo-box styles.

IMO, most of these shared styles are totally unhelpful with the light combo-box, maybe except for the `small` icon and the simple state switches. Maybe move back to the regular combo-box for now? Otherwise needs rework.

---

*[vaadin-combo-box.html, line 192 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8G7-oGottgoeaJWI5:-KK8G7-oGottgoeaJWI6:73605336) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7bd54ea4cd62ae6b7fbabf6d2b4e844904c749fd/vaadin-combo-box.html#L192)):*
> ```HTML
>     },
> 
>     // document.activeElement doesn't return elements from inside shadow root
> ```

This comment line is now lost.

---

*[vaadin-overlay-behavior.html, line 93 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK83YsStd84b4IEflin:-KK83YsTvm4HDBPaShyk:-954297247) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-overlay-behavior.html#L93)):*
> ```HTML
>     _isPositionFixed: function(element) {
>       var parentElement = element.parentElement;
>       if (!parentElement || parentElement.localName === 'content') {
> ```

After thinking twice about this condition, I’m not sure.

Do we jump over strait to the `x-foo` host, skipping the `some-containter` check, in case when the DOM tree is like this?
```html
<dom-module id="x-foo">
  <template>
    <some-container style="position:fixed">
      <content>
        <checked-element></checked-element>
      </content>
    </some-container>
  </template>
</dom-module>
```

---

*[demo/custom-input.html, line 43 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8BCqnbF2D2QOLBQlH:-KK8BCqopX-lwEUwLYGV:578270021) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/demo/custom-input.html#L43)):*
> ```HTML
>       <style is="custom-style">
>         vaadin-combo-box paper-icon-button {
>           color: rgba(0, 0, 0, .38);
> ```

`paper-icon-button.clear-button` should receive the same styles by default, right? Then why do we have unnecessary styles here?

---

*[demo/custom-input.html, line 111 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8C4du_X57Y0YEeWov:-KK8C4dvsTm2zmfrzzqR:-1823072130) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/demo/custom-input.html#L111)):*
> ```HTML
>       <style>
>         /* Hide the clear button if there is no value to clear. */
>         vaadin-combo-box-light:not([has-value]) .clear-button {
> ```

Same question about the unnecessary style here.

---

*[docs/vaadin-combo-box-customization.adoc, line 2 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8COVuk2BxooRXsei6:-KK8COVvt2lBJtqOvhvn:-171148535) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/docs/vaadin-combo-box-customization.adoc#L2)):*
> ```AsciiDoc
> ---
> title: Customization
> ```

Would be better to have a bit more specific title, e.g.: `Customizing the Input Area`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 19 of 20 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 39 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7ig-Bg2wpLdrRTsNI:-KK8KtSQnPWitNPZfOjW:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L39)):*
> Add a default value here, maybe? Just to make it more explicit and clear.
</details>
Done.

---

*[vaadin-combo-box-behavior.html, line 86 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7j2sqsNz5mU6wrJ4o:-KK8KWgFDB6GsCzzxopW:-135602928) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L86)):*
> This became public. Feels like it needs an APIdoc comment then.
</details>
Hmm.. it became private, not public.

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KK8KEqsDUJsCWxuhBA-:-2083831195) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
> It's a group header for the next few methods. Please bring it back.
</details>
I personally find that kind of grouping comments unnecessary and easily outdated as the code evolves. I would keep it out. :smile: 

---

*[vaadin-combo-box.html, line 192 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8G7-oGottgoeaJWI5:-KK8MK99e6pMhDGSTp_0:495399959) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7bd54ea4cd62ae6b7fbabf6d2b4e844904c749fd/vaadin-combo-box.html#L192)):*
> This comment line is now lost.
</details>
Added this comment back slightly re-formatted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 18 of 20 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[docs/vaadin-combo-box-customization.adoc, line 2 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8COVuk2BxooRXsei6:-KK8Qwx9hBUgxaidd_F_:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/docs/vaadin-combo-box-customization.adoc#L2)):*
> Would be better to have a bit more specific title, e.g.: `Customizing the Input Area`.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Reviewed 1 of 2 files at r7.
Review status: 19 of 20 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KK8X2oZe_8MV3yuNVyT:1924499108) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> I personally find that kind of grouping comments unnecessary and easily outdated as the code evolves. I would keep it out. :smile: 
</details>
Well, it was absolutely necessary in the combo-box, and for the same reasons it is necessary here.

A while ago we had a ton of methods for spread quite randomly. After some refactoring, we end up with these grouping headers and sections in the combo-box. Now we just extract almost everything into a behavior. Why should we remove section headers at this time?

Why have you decided to remove just one of the headers, leaving the others in place? We also have “Keyboard navigation”, “Filtering and items handling” grouping headers here.

As the code evolves, the correct sectioning and order should be maintained, otherwise it becomes messy and hard to follow. I don't want the mess to be back here. Let’s rather have a header back, please.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Review status: 19 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks broke.

---

*[vaadin-combo-box-behavior.html, line 86 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7j2sqsNz5mU6wrJ4o:-KK8_KBbZ_qETaCKQl62:1212512615) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L86)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> Hmm.. it became private, not public.
</details>
Don’t mind, I commented the outdated revision.

I would rather make it back public, though.

The combo-box would anyways have the public `inputElement` property inherited from `PaperInputBehavior`. It doesn‘t make sense to duplicate it as a private property here, but rather would be meaningful to overload the `inputElement` from the behavior.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 18 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 86 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7j2sqsNz5mU6wrJ4o:-KK8avgXhlPgBM_Xk11v:1628457787) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L86)):*
> Don’t mind, I commented the outdated revision.
> 
> I would rather make it back public, though.
> 
> The combo-box would anyways have the public `inputElement` property inherited from `PaperInputBehavior`. It doesn‘t make sense to duplicate it as a private property here, but rather would be meaningful to overload the `inputElement` from the behavior.
</details>
You're right. I originally missed the fact that it was inherited from `PaperInputBehavior`. I'll update the code.

---

*[vaadin-combo-box-behavior.html, line 257 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq6PCYr8yKqxE0syJ6:-KK8apnS46eykW4pRgfH:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L257)):*
> Well, it was absolutely necessary in the combo-box, and for the same reasons it is necessary here.
> 
> A while ago we had a ton of methods for spread quite randomly. After some refactoring, we end up with these grouping headers and sections in the combo-box. Now we just extract almost everything into a behavior. Why should we remove section headers at this time?
> 
> Why have you decided to remove just one of the headers, leaving the others in place? We also have “Keyboard navigation”, “Filtering and items handling” grouping headers here.
> 
> As the code evolves, the correct sectioning and order should be maintained, otherwise it becomes messy and hard to follow. I don't want the mess to be back here. Let’s rather have a header back, please.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 14 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 86 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK7j2sqsNz5mU6wrJ4o:-KK8exVaWkV10XFFWxyG:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L86)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> You're right. I originally missed the fact that it was inherited from `PaperInputBehavior`. I'll update the code.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

manolo:
Review status: 14 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KK8dIqONJ1Na1FhSPuf:633130052) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
> Disagree. The `String` case is different from the `Object` here. Strings are valid items as they are, they are not treated as like any objects. Strings should be mentioned explicitly here.
</details>
We need this to be correctly documented in order to generate correct signatures in java. I think it's not worth to produce two methods when actually one with the Object signature would be enough. 

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KK8el7dXA4VS882oEzw:54746505) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
> The method was public in the `vaadin-combo-box` API already. It’s not specific for a particular input element, so it should be in `vaadin-combo-box-light` as well.
> 
> Though we do export the behavior, you can’t call its methods directly. It doesn’t work like this. We should rather consider the behavior methods / props / events as the elements’s API, not the behavior's API.
> 
> This way it makes a perfect sense to extract the identical combo-box methods to the combo-box-behavior, both public and private ones.
</details>
Well, it's not clear to me when we should make one method public or private here.  Why does the user need to use cancel? and not _clear?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 19 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box.html, line 184 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqEZw4pvomrxbtM4uN:-KK8idgOMrjxQtJspvl0:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box.html#L184)):*
> I mean, “*Note* that the combo-box...”
</details>
Done.

---

*[vaadin-overlay-behavior.html, line 93 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK83YsStd84b4IEflin:-KK8ii-43Ceqc0DMztuD:-608747647) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-overlay-behavior.html#L93)):*
> After thinking twice about this condition, I’m not sure.
> 
> Do we jump over strait to the `x-foo` host, skipping the `some-containter` check, in case when the DOM tree is like this?
> ```html
> <dom-module id="x-foo">
>   <template>
>     <some-container style="position:fixed">
>       <content>
>         <checked-element></checked-element>
>       </content>
>     </some-container>
>   </template>
> </dom-module>
> ```
</details>
You're right. This might need some more thought.

---

*[demo/custom-input.html, line 43 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8BCqnbF2D2QOLBQlH:-KK8j6-CEzlXQx8jC6sA:-1773345508) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/demo/custom-input.html#L43)):*
> `paper-icon-button.clear-button` should receive the same styles by default, right? Then why do we have unnecessary styles here?
</details>
True, these are no longer needed and were left from the previous implementation. Removed.

---

*[demo/custom-input.html, line 111 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK8C4du_X57Y0YEeWov:-KK8jAoaBAOJ5JlehcEA:-1860574884) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/demo/custom-input.html#L111)):*
> Same question about the unnecessary style here.
</details>
Removed also this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 16 of 20 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-shared-styles.html, line 42 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK87rVBNWyW2jfjq47E:-KK8l2QcaviFsiYqZ1Kr:-899875444) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-shared-styles.html#L42)):*
> These absolute positioning styles for the clear and the toggle buttons require complementary equal padding for the input element. This condition is fulfilled for the regular combo-box, but not for the light one.
> 
> For example, if someone use `<paper-input-button class="toggle-button">` with the light combo-box, they would have a potential overlap of the icon and the input value. To resolve this, they have to figure out and extract the correct solution from the combo-box styles.
> 
> IMO, most of these shared styles are totally unhelpful with the light combo-box, maybe except for the `small` icon and the simple state switches. Maybe move back to the regular combo-box for now? Otherwise needs rework.
</details>
You're right. Moved most of the styles back to `vaadin-combo-box`.

---

*[vaadin-combo-box-shared-styles.html, line 73 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KK80zRdqnicVIfn1DAQ:-KK8l29f5I_n-pD50iac:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-shared-styles.html#L73)):*
> Looks like all the styles above that are scoped to `paper-input-container` wouldn't apply for the combo-box-light. They are limited to the paper-input-container in the shadow DOM of the element, which is never true in case of the light combo-box.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
:lgtm:

<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> #### Introduce input customization feature
> 
> Fixes #198 
> 
> 
> 
> **IMPORTANT! Please don't merge this PR just yet. We need to rebase, clean up and squash the commit history before merging.**
> 
> 
> 
> This is quite a big change and therefore I tried to list all the main points here.
> 
> 
> 
> Ground work
> 
> ----
> 
> * Our `iron-icon` + `paper-ripple` combo is now replaced with `paper-icon-button`.
> 
> * Following attributes are now dynamically added/removed to the host element to enable styles based on the component state.
> 
>   * `opened`
> 
>   * `has-value`
> 
>   * `readonly`
> 
>   * `disabled`
> 
> 
> 
> Customization feature
> 
> ----
> 
> * Replacing the `paper-icon-buttons` can be done by adding child elements with class names of `toggle-button` and/or `clear-button`.
> 
> * Two new class names are exposed for styling.
> 
>   * `small` – makes the icon slightly smaller (applied to the default clear icon, only affects a `paper-icon-button` element).
> 
>   * `rotate-on-open` – makes the button rotate with a transition when dropdown is opened (applied to the default toggle icon).
> 
> 
> 
> 
> 
> Example.
> 
> ```html
> 
> <vaadin-combo-box>
> 
>   <paper-icon-button icon="remove-circle" class="clear-button"></paper-icon-button>
> 
>   <paper-icon-button icon="search" class="toggle-button"></paper-icon-button>
> 
> </vaadin-combo-box>
> 
> ```
> 
> 
> 
> * For replacing the whole input we introduced a new element called `vaadin-combo-box-light`.
> 
>   * This split will avoid unnecessary imports when using a customized version.
> 
>   * All logic from `vaadin-combo-box` was refactored to new behavior called `vaadin-combo-box-behavior`.
> 
>   * The `vaadin-combo-box-light` expects to have a child element matching the selector `input[is="iron-input"],paper-input,.paper-input-input,.input`.
> 
>   * The input element must have some two-way data-bindable property for the current value.
> 
>   * By default the value property is expected to be `bind-value` (to be compatible with `iron-input`).
> 
>   * Value property name can be changed with the `attrForValue` property.
> 
> 
> 
> Examples.
> 
> ```html
> 
> <vaadin-combo-box-light>
> 
>   <input is="iron-input">
> 
> </vaadin-combo-box-light>
> 
> 
> 
> <vaadin-combo-box-light attr-for-value="value">
> 
>   <paper-input label="Elements" class="input">
> 
>     <iron-icon icon="toll" prefix></iron-icon>
> 
>     <paper-button suffix class="clear-button">Clear</paper-button>
> 
>     <paper-button suffix class="toggle-button">Toggle</paper-button>
> 
>   </paper-input>
> 
> </vaadin-combo-box-light>
> 
> ```
> 
> 
> 
</details>


---

Reviewed 4 of 4 files at r10.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 14 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJq92yvyPo4YHmeDSKm:-KK8lrs_WZcdhhYIrDNB:1440999536) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L14)):*
> We need this to be correctly documented in order to generate correct signatures in java. I think it's not worth to produce two methods when actually one with the Object signature would be enough. 
</details>
Isn't it easier to understand if there are two notations? Would prefer it, if it’s not harmful.

One single `Array<Object>` notation makes me think that I need to make some standart properties on top of my strings to make them Object-compliant. This is confusing, actually you can just pass strings and combo-box would work with plain strings just fine, no need for attaching any sub-properties.

---

*[vaadin-combo-box-behavior.html, line 281 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KJqAwzTewk11lRt9rXG:-KK8orw6bQIxCvA_dtey:-1280948750) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/1981bbbb52b406ec187b24b042c47d965bb9904a/vaadin-combo-box-behavior.html#L281)):*
> Well, it's not clear to me when we should make one method public or private here.  Why does the user need to use cancel? and not _clear?
</details>
Interesting question, but not in appropriate time. Regardless of the answer, we can’t make this method private in a non-major release.

I think, the reason is that previously developers didn't have direct access to the input value, only to the combo-box value. This method covers the common use case, when you might need the access in order to stop filtering, discard the user-entered text and return the label of the selected item (or empty). Without the method, it's awkward to do the same.

Unlike the `cancel()` method, the `_clear()` is easy and straitforward to do with other existing public API’s. Just set the value property to empty string and it’s done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-:-KK8xKx5_LaEeCTRN8w8:1417088721)*

tehapo:
Squashed and rebased. :racehorse: 

---

Review status: 16 of 20 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-:-KKCe3_flwVDpPzmipHV:1409965636)*

Saulis:
Reviewed 1 of 20 files at r1, 1 of 1 files at r2, 2 of 11 files at r3, 1 of 5 files at r9, 2 of 4 files at r11.
Review status: 18 of 20 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 90 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCfeXkPFsmSwhMFFsO:-KKCfeXlu0Yih2CnzSDO:-1974961889) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-behavior.html#L90)):*
> ```HTML
>       // PaperInputBehavior, which already has "inputElement" defined and we
>       // don't want to have two separate properties.
>       inputElement: Object,
> ```

Doesn't this property conflict with the `PaperInputBehavior` then? Which one overrides which?

---

*[vaadin-combo-box-light.html, line 105 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCkJMI_LNW4ikTVVKz:-KKCkJMI_LNW4ikTVVL-:-2112719341) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-light.html#L105)):*
> ```HTML
>       this._clearElement = Polymer.dom(this).querySelector('.clear-button');
> 
>       this._revertInputValue();
> ```

What's this `_revertInputValue()` call for?  Set the possibly selected value to the input field?

---

*[vaadin-combo-box-light.html, line 106 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCifta4jyr-ulNvOli:-KKCifta4jyr-ulNvOlj:-1191585813) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-light.html#L106)):*
> ```HTML
> 
>       this._revertInputValue();
>       this.listen(this, 'down', '_preventDefault');
> ```

Missing a `detached` to `unlisten` all these listeners?

---

*[vaadin-combo-box.html, line 219 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCnT5zPZ0B2dBQZD2d:-KKCnT5zPZ0B2dBQZD2e:-888006343) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box.html#L219)):*
> ```HTML
>     behaviors: [
>       vaadin.elements.combobox.ComboBoxBehavior,
>       Polymer.PaperInputBehavior
> ```

Adding `PaperInputBehavior` smells a bit like a feature creep. Also, one of the main reasons for providing the custom input feature was to avoid tighter coupling between `vaadin-combo-box` and `paper-input` and its API. Not sure what to think about this...

---

*[test/selecting-items.html, line 58 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCoittQbWjzFCi7Ckr:-KKCoittQbWjzFCi7Cks:1660672) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/test/selecting-items.html#L58)):*
> ```HTML
>     });
> 
>     it('should fire value change event', function() {
> ```

Should still fire only one `value-changed` ? Right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 18 of 20 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 90 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCfeXkPFsmSwhMFFsO:-KKDS5s3ToLdVwCyzsZp:115714883) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-behavior.html#L90)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Doesn't this property conflict with the `PaperInputBehavior` then? Which one overrides which?
</details>
`PaperInputBehavior` is now removed.

---

*[vaadin-combo-box.html, line 219 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCnT5zPZ0B2dBQZD2d:-KKDSBqkPHvYsNN6uL_u:115714883) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box.html#L219)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Adding `PaperInputBehavior` smells a bit like a feature creep. Also, one of the main reasons for providing the custom input feature was to avoid tighter coupling between `vaadin-combo-box` and `paper-input` and its API. Not sure what to think about this...
</details>
`PaperInputBehavior` is now removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 17 of 20 files reviewed at latest revision, 7 unresolved discussions.

---

*[test/selecting-items.html, line 58 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCoittQbWjzFCi7Ckr:-KKDTM602OaZsXg3bRy-:-518473674) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/test/selecting-items.html#L58)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Should still fire only one `value-changed` ? Right?
</details>
Reverted this change. Let's see if all tests still pass.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 16 of 20 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-combo-box-light.html, line 105 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCkJMI_LNW4ikTVVKz:-KKDUcwrd4dIpcWV9LBV:-1834165927) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-light.html#L105)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> What's this `_revertInputValue()` call for?  Set the possibly selected value to the input field?
</details>
@platosha can you remember why you added this call to this `attached` method?

---

*[vaadin-combo-box-light.html, line 106 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCifta4jyr-ulNvOli:-KKDUcUr3HOZGyp1xxAk:-499146009) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-light.html#L106)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Missing a `detached` to `unlisten` all these listeners?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

platosha:
Review status: 16 of 20 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-combo-box-light.html, line 105 \[r11\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKCkJMI_LNW4ikTVVKz:-KKDV87wtpTQ61jQSjeI:977439973) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/7871ebf112b786060cd670a6ed993e4704aa0840/vaadin-combo-box-light.html#L105)):*
<details><summary><i>Previously, tehapo (Teemu Pöntelin) wrote…</i></summary>
> @platosha can you remember why you added this call to this `attached` method?
</details>
Yes, exactly for setting the possibly selected value to the input field.

---

*[vaadin-combo-box-light.html, line 106 \[r13\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKDVS7P_cwAJRboQFuX:-KKDVS7Qjln8a5vaBVTq:1381575792) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b044ab62d65e8f066a6fa7449d7414017d32359a/vaadin-combo-box-light.html#L106)):*
> ```HTML
> 
>       this._revertInputValue();
>       this.listen(this, 'down', '_preventDefault');
> ```

This line with the pair `unlisten` could be refactored into `listeners: { 'down': '_preventDefault' },`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

tehapo:
Review status: 16 of 20 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-combo-box-light.html, line 106 \[r13\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-KKDVS7P_cwAJRboQFuX:-KKHhsU7iiNl23sA_wzS:395572573) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b044ab62d65e8f066a6fa7449d7414017d32359a/vaadin-combo-box-light.html#L106)):*
> This line with the pair `unlisten` could be refactored into `listeners: { 'down': '_preventDefault' },`.
</details>
You're right.  That one listener could be moved to `listeners` property, but I would rather keep it as is to have all the listeners in the same place.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254)*

Saulis:
:lgtm:

---

Reviewed 4 of 20 files at r1, 1 of 11 files at r3, 1 of 5 files at r9, 3 of 4 files at r10, 2 of 2 files at r12, 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/254#-:-KKHjp-aqn4xVGBZX4SM:1417088721)*

PR #268, opened 2016-07-03
Create custom iconset

web-padawan:
Actually `vaadin-combo-box` only uses a few icons, so it would be nice not to import all of the `iron-icons.html`. I created a custom iconset for that.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/268)


manolo:

---

*[vaadin-combo-box-icons.html, line 1 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/268#-KM3RsMFUPqhAWLvloOj:-KM3RsMFUPqhAWLvloOk:-112391561) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/10ba66fc6df69d81aa8ff11b4f871176744d0ac1/vaadin-combo-box-icons.html#L1)):*
> ```HTML
> <link rel="import" href="../iron-iconset-svg/iron-iconset-svg.html">
> ```

I'm not have a definitive preference, but I feel that since this file is really small, I would rather to include in the vaadin-combo-box so as we reduce the one request penalty. I know that most people would vulcanise the output, but everything counts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/268)*

tomivirkki:
The PR itself looks fine to me, but the change is a bit of a tradeoff: It adds (very little) weight to projects that already import iron-icons but loses (some more) weight from projects that don't.

All Vaadin Core Elements should be aligned so we'd also require the same changes for our other elements that use iron-icons (date picker, upload).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/268#-:-KM3SaTKYBgbiKPZyxeg:-273243921)*

web-padawan:
@manolo I updated the PR, now iconset is included to `vaadin-combo-box` as you suggested.

web-padawan:
@tomivirkki did you mean creating an extra repo for all Vaadin Core Elements containing those icons? Or encapsulating icons within each separate component?
Actually, I only found a few icons used by upload and datepicker as well.

Saulis:
@web-padawan this approach works fine, the icons used in date-picker and upload are different so I don't think there would be benefit bundling them in the same iconset. 

If you could rebase this branch, I'd be happy to merge it.

Also, would be super awesome if you could also send us a PR for date-picker and upload for similar changes 😄

web-padawan:
@Saulis thanks! It would be nice. I rebased the branch.
Sure, I'll send a PR soon for date-picker and upload as well :)

Saulis:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/268#-:-KN_pSKENlyiXYXCJUVs:1417088721)*

Saulis:
@web-padawan hey there! We actually already got an PR for the upload coming in 😄  

Also, I realised that `iron-icons` uses BSD license which should be included with the icons – so I think we after all need to have the icons in a separate file which also contain both Apache-2.0 license for Vaadin and the original BSD license. 

I'll fix that for combo-box, but please take this into account if you have the time to make a PR for date-picker.

web-padawan:
@Saulis got it, thanks. I'll create a PR for date-picker this weekend.

PR #281, opened 2016-07-25
Set inputElement to read-only to prevent angular2-polymer overriding …

### LINKED ISSUE #280: this.inputElement is undefined throw
While using vaadin-combo-box in my angular 2 project, some options not shown in first attempt, when review in firbug it throw error

TypeError: this.inputElement is undefined
if (!this.inputElement.focused) { ....

I tried my best to investigate why it happens but failed. can someone know why it causes. i am using firefox and chrome.

--
Screen of output (only display few items even there are more items in first attemp)

issue01

Next screen shows error in firebug.

issue02

Saulis:
Fixes #280

Discussed also the options of setting the property with notify:true or making
it private, but decided to just make it readOnly as originally in the
PaperInputBehavior the property only has a getter.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/281)

### DIFF
--- a/test/angular2.html
+++ b/test/angular2.html
@@ -126,6 +126,10 @@
           it('should not reflect invalid state to combobox initially', function() {
             expect(combobox.invalid).to.be.false;
           });
+
+          it('should have a value set in inputElement', function() {
+            expect(combobox.inputElement).to.not.be.empty;
+          });
         });
 
         describe('after a value change', function() {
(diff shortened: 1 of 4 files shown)


platosha:
:lgtm:

We should probably fix the Date Picker in the same way.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/281#-:-KNX08Aw_Hk0rKmkopX3:-1712668710)*

Saulis:
`vaadin-date-picker` actually seems to have an getter for `inputElement`

PR #282, opened 2016-07-25
Fire change event on value change

### LINKED ISSUE #271: Implement 'change' event
- 'change' is the standard name in w3c for value changes:  http://www.w3schools.com/jsref/event_onchange.asp `The onchange event occurs when the value of an element has been changed.` 
- Hummingbird assumes that all Inputs update value property and fire a 'change' event
- Polymer is firing 'change' in inputs, text-area etc.

Additionally in the combo-box the change event coming from the underlying input is leaking out to the element, and it's observed before the actual value is updated which could be confusing for the user

Saulis:
Fixes #271

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/282)

### DIFF
--- a/docs/vaadin-combo-box-value.adoc
+++ b/docs/vaadin-combo-box-value.adoc
@@ -44,7 +44,11 @@ Since there are two selection properties, namely [propertyname]#value# and [prop
 an event is fired for each. [propertyname]#value-changed# event contains the selection as a String value,
 whereas [propertyname]#selected-item-changed# contains the selection as an Object.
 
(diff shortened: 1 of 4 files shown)


platosha:
:lgtm:

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/282#-:-KNaAO_VAzcqt4j-TSSX:1417088721)*

PR #300, opened 2016-09-05
feat(combo): Handle multiple properties in combo box label

JulienNexonia:
This branch allows to handle both single and multiple properties in combo box label.
Very interested in using "custom properties for item label and value", I added multiple label properties support.

As you can see in the demo file, you can now use several properties for the label of the combo as follows : 
`item-label-path="name"`

You can even specify a separator between the desired values as follows : 
`item-label-separator=" - "`

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/300)

### DIFF
--- a/demo/object-items.html
+++ b/demo/object-items.html
@@ -230,6 +230,45 @@ layout: page
     </code-example>
   </section>
 
+
+  <section>
+    <h3>Custom Properties for Item Label and Value</h3>
+    <p>
+      You can override the default property paths by defining <code>item-label-path</code> and <code>item-value-path</code> properties.
+      <br/>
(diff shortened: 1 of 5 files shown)


jouni:
Hi, and thanks for the PR!

This is a nice feature, but we believe this use case is better covered by the upcoming custom item template feature (#165). So this implementation is not something we are willing to merge, unfortunately :(

We really appreciate all the help we get from our users/community, and don’t like to turn down PRs like this. To avoid similar situations in the future, it would be great if you would open up an issue for any new feature/bug you are planning to work on (search for existing ones first), so we can coordinate and avoid wasted effort.

PR #305, opened 2016-09-13
Fix vaadin-combo-box compatibility with iron-list 1.3.9

### LINKED ISSUE #304: Regression in overlay positioning after iron-list update
After updating `iron-list` to 1.3.9 the overlay positioning is not correct when having the `vaadin-combo-box` near the bottom of the viewport. Seems to be corrected after closing and reopening.

iron-list-update

tehapo:
Fixes #304

The latest `iron-list` (v1.3.9) introduced more asynchrony to the initial rendering in order to improve performance. This broke a lot of our tests and also introduced two bugs (see #304).

This PR attempts to resolve the problems by adding an asynchronous `notifyResize` call to the `attached` callback of the overlay also wrapping `_scrollIntoView` call into an `async` callback. Not very nice things to do, but resolves the two bugs.

Also tests have been modified to work with these changes and the asynchrony introduced by `iron-list` update.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/305)

### DIFF
--- a/test/aria.html
+++ b/test/aria.html
@@ -86,15 +86,18 @@
         arrowDown();
       });
 
-      it('should set selection aria attributes when focusing an item', function() {
+      it('should set selection aria attributes when focusing an item', function(done) {
         getSelector().selectedItem = 'foo';
         arrowDown();
 
-        expect(getItemElement(0).getAttribute('role')).to.equal('option');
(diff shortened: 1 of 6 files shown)


tehapo:
Abandoning this PR in favor of #307.

PR #326, opened 2016-10-16
Custom dom elements for items. Fixes #106

### LINKED ISSUE #106: Highlight matches in the filtered list

manolo:
- User can customize item content by setting a dom
  element which is added to the item object.
- Alternative approach to #200 
- Note that this could be done allowing the user to pass an html string, but we want to protect the component against XSS vulnerability.
- Item templates #165  is out of the scope of this PR. 

For reviewers:
This PR needs the public-filter branch

---

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/326)


platosha:
:thumbsdown: Here we’re mutating the items, and even worse, store DOM elements right in the items. This is a huge disadvantage, IMO. Blending DOM and data together is an anti-pattern and should not be encouraged.

The items mutation will affect all the use cases of sharing the items array for the other purposes, like displaying them elsewhere, caching in the browser, storing in the DB backend... Even two vaadin-combo-box elements on the same page bound to the same items array might clutter within each other.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUHXcvpQGceO2cuEVa8:b-w9xcg)*

manolo:
Well, I absolutly agree that modifying the items might be a bad idea.

But thinking a little more it happens that 
1.- the component does not modifies anything, it just uses it. Combo box is agnostic to the item source. Its API only demands a configuration object with a field `value` and optionally `label` and `element`. 
2.- it's the user's responsibility  to provide the appropriate item configuration, whether he takes it from a db or ajax and whether he reuses the items or just creates new ones with the info needed by the component should be out of the scope of the component.

Do you have other ideas about how to provide a dom structure per item.

I would suggest that the user could provide a renderer, or otherwise we have to go with custom templates so as the user could eventually code a custom element scoped with the item object

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KULVuZQsKSJRpU5mn7b:b4ima66)*

platosha:
Well, as long as the user is required to modify the data to provide DOM elements in there, this is unfriendly. It’s like a requirement for user to mix M and V of MVC for using this feature.

I believe, templates are the way to go. They are declarative, they are separate from data, they are specific for a particular combo-box and do not require data mutations, will not clutter.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUMTtEal_QgEFajl5bS:b-n5ennw)*

Saulis:
As a general comment, I think we should encapsulate the highlighting logic into a custom element something like:

```html
 <text-highlighter match="foo" string="foobar"></text-highlighter>
```

1) Provides reusability and encapsulation
2) Can provide clearly targeted css properties and mixins to customize the highlight styles
3) Directly compatible usage with the upcoming custom item templates feature: 
```html
<template>
  <text-highlighter match="[[filter]]" string="[[item.value]]"></text-highlighter>
</template>
```

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUWVnFSPSWjDGla2Yag:b824vir)*

platosha:
@saulis nice suggestion!

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUXHfhA0UKB0sP3meZf:b-794p4g)*

manolo:
How do you declarative match  `<b>F</b>inni<b>s</b>h` when the filter is `FS` in that custom element?

I think the best is that the user provides whatever he wants in the template, that obviously includes that he can creates his own custom element. For instance to show flags  in a list of countries and select the appropriate letters based on any criteria.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUXJtxol_NliVVZTAjL:b-yyu23p)*

platosha:
@manolo I think, the supplied text highlighter should handle a simple common case, for instance, highlight all occurrences of a given substring. Then users can implement a custom highlighter element if they need to.

If we support regexp in the text highligher, then computed bindings could be used to solve your case.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUXP4sdurD2FLX2NXXK:bsiovwx)*

manolo:
So, to make things easier and more coherent the `<text-highlighter>` could be the component  that handles the search as well. 

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/326#-:-KUXmGmgFzxlIt98cBAj:b3d31x)*

Saulis:
Yeah the element would definitely use regexp which could be customized

Saulis:
Just as food for thought, here's a proto I made: http://jsbin.com/dufilu/edit?html,output

Issue #259, opened 2016-06-21
Custom filtering


jouni:
As a developer, I want to control how the filtering of the items works. I want control over what keys in the item objects the filtering matches and how that matching is done ("begins with" rather than "contains" for example).

jouni:
Related: #106

PR #327, opened 2016-10-16
Public filter API. Fixes #284

### LINKED ISSUE #284: Expose input-element-value to facilitate AJAX requests
At the moment, I would like to do something like this:

```
<iron-ajax auto={{searchName}} url="/stores/polymer?nameStarts=[[searchName]]" last-response={{ajaxResponse}}></iron-ajax>
<vaadin-combo-box id="VCC" input-element-value="{{searchName}}" label="Language" class="languages" items="[[ajaxResponse]]"></vaadin-combo-box>
```

Basically, the AJAX request is initiated when something is typed. iron-ajax makes the request to the server, and `{{ajaxResponse}}` is bound to the last response. That bound variable is fed to the wonderful vaadin-combo-box, and voila' -- only the "legal" options are available. Nirvana of auto-completion is reached.

But! vaadin-combo-box doesn't expose input-element-value. The change is trivial: just add:

```
  inputElementValue: {
    type: String,
    notify: true
  },
```

And then:

```
 /**
 *  Filtering and items handling
 */

_inputValueChanged: function(e) {

  // Expose inputElementValue
  this.inputElementValue = this._inputElementValue;
```

To `vaadin-combo-box-behavior.html`. These two changes basically mean that the code above works. I tried a million possible ways to avoid changing vaadin-checkbox, but no -- if the change isn't exposed, it's just not going to work!

For the sake of making AJAX autocompletion work, would you accept a pull request with the changes above?

Another option would be to rename `_inputElementValue` into `inputElementValue`. However, this would be less than ideal. Really, in terms of facilitating AJAX requests, we should only change `inputElementValue` if there has been an `input` event on the input field. So, in terms of behaviour in terms of making an AJAX request, the way I proposed is _perfect_.

Maybe inputElementValue isn't _actually_ the best name for it. It should be something like `keyedInValue` or... something that clarifies that it's the result of user typing (hence the auto-completion query).

The only way I have to go about it is to write a decorator widget that mirrors _every_ possible input of `vaadin-combo-box.html`, and uses an extra behaviour that overwrites `_inputValueChanged`. But that really is way less than ideal...! Alternatively, I can clone `vaadin-combo-box.html` so that it redefines `_inputValueChanged` but even then...

It would be awesome if `vaadin-combo-box` was usable with AJAX-loaded data out of the box. Especially since the modifications would be so minimal...

Thanks for listening!

manolo:
- making _filter property public so as the user can
  bind to it and set the items using custom code.
- adding a filter-disable property to disable internal
  filtering code.

For reviewers:
Merge this into update-demos

---

This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/327)


platosha:
I wonder is `filterDisabled` property a future-proof solution in the scope of the custom filtering feature (#259)? I mean:

- Can we resolve #259 with it?
- If not, would it still be needed having a custom filtering feature after #259 is solved?

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[test/filtering.html, line 217 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUHbrcWVMjF7HxfAMt5:-KUHbrcWVMjF7HxfAMt6:bi35xxy) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/test/filtering.html#L217)):*
> ```HTML
>         setInputValue('a');
>         expect(getFilteredItems()).to.eql([ 'foo', 'bar', 'baz' ]);
>       });
> ```

Please test that the `filter` property change notifications are fired.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUHaSrUPzrPTcigHJis:b2huda7)*

manolo:
something like  `filter-strategy = none|contains-word|contains-characters|begins-with-word|ends-with-word` would be enough user cases. Suggestions for the property name and values ? 

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[test/filtering.html, line 217 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUHbrcWVMjF7HxfAMt5:-KULS5xd_ltH5uuqTlJB:bg6nmyz) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/test/filtering.html#L217)):*
> Please test that the `filter` property change notifications are fired.
</details>
That is guaranteed by the polymer framework that happens. Or do you mean check that no one removes the notification accidentally ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KULRQNTCgI1jTRJEGMP:b-kkvc00)*

platosha:
What about:
```typescript
/**
 * String to use in a RegExp constructor
 * Set to empty to disable filtering. Use compound binding binding with [[filter]]
 * to declaratively define a custom filter matching expression.
 */
filterExpression: string = `^${filter}`;

// Example usage in the _filterItems callback
_filterItems: function(array: string[], filterExpression: string) {
  if (!filterExpression) return array;
  
  const regExp = new RegExp(filterExpression, "i");
  return array.filter(regExp.test.bind(regExp));
}
```

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUMHqzDL-hxVUdDZciN:b1xw4m2)*

manolo:
That not work in the case of wanting to match characters that the user types not consecutive. For instance the user types `sai` and it should select `SpAIn` (Upercase is the matching letters). 
The expression changes any time the user types something  
`s -> .*s.*`  
`sa -> .*s.*a.*`
`sai -> .*s.*a.*i.*`

I think a set of preconfigured filtering functions like I said should cover many cases. Otherwise the user has to pass a matching function so as the regular expression changes with the input.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUNh7QYGMpXkTRNkYv9:bv74656)*

platosha:
Well, that’s true, this particular case probably can’t be achieved with compound bindings, but still possible to implement, and pretty easy with computed bindings.

Back to the original question. Whatever we decide for the custom filtering API, it will likely cover the `filterDisabled` case as well. Therefore, we should probably not include `filterDisabled` in this PR. Will the filter property still be usable without disabling or customising the internal filtering?

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUQxgDBipn93EdIrobo:b-d781vn)*

platosha:
Or am I wrong and this is the other way around: `filterDisabled` + `filters` solves #259 as well, by allowing users to implement a custom filtering outside the combo box, se we don’t need to invent anything on top of it?

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[test/filtering.html, line 217 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUHbrcWVMjF7HxfAMt5:-KURsqFdA6XzXnQMWF-U:b-1vijnd) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/test/filtering.html#L217)):*
> That is guaranteed by the polymer framework that happens. Or do you mean check that no one removes the notification accidentally ? 
</details>
Exactly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KURsR5ZeEFjyc67XB3r:b-k4nd0a)*

Saulis:
I wouldn't touch the filtering itself in this PR, instead would just instruct to use the same filtering logic remotely as we do for now. If this isn't good enough, then we should design and implement #259 before going further with this.

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 464 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUWU3dXY8aIJnH2qJEA:-KUWU3dXY8aIJnH2qJEB:b-9oseeg) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/vaadin-combo-box-behavior.html#L464)):*
> ```HTML
>       this.$.overlay.notifyPath('_items', undefined);
> 
>       if (!items) {
> ```

Why is this here? So what happens if `items` is changed from a populated array to an empty/undefined array? 

---

*[vaadin-combo-box-behavior.html, line 478 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUWUBPSLqLOVD6hkrKS:-KUWUBPSLqLOVD6hkrKT:b5krhx) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/vaadin-combo-box-behavior.html#L478)):*
> ```HTML
>       this.$.overlay.notifyResize();
> 
>       // async needed to reposition correctly
> ```

Why is this needed now? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUWV7kNT-d3tHRzFl4R:b-ogzd8c)*

manolo:
You cannot force the user to use our simple implementation of filtering. Just a way for disabling it gives the user full control on filtering as @platosha said. 

Couple of cases that the internal disable filtering  covers:

1.- the user don't want at all any filtering. He wants the combo-box behave like html selects.
2.- the remote filtering has other criteria. That the case for instance of the example in the demo with github where it returns relevant cases for it that does fits the internal filtering
3.-  The usage of advanced engines like Solr that supports fuzzy search, phonetic search, sentiment search, etc. That cases does not work if we filter with our trivial fragment matching.

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 464 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUWU3dXY8aIJnH2qJEA:-KUX9jNEiNiXxP8003D4:b-qkyprr) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/vaadin-combo-box-behavior.html#L464)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Why is this here? So what happens if `items` is changed from a populated array to an empty/undefined array? 
</details>
When the user erases the input with back character, etc

---

*[vaadin-combo-box-behavior.html, line 478 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUWUBPSLqLOVD6hkrKS:-KUX9r8usi1tXCOoo4lP:b-omyw39) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/vaadin-combo-box-behavior.html#L478)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary>
> Why is this needed now? 
</details>
Otherwise it not works, have not dive so much on it, but if you change items when the overlay is open it does not resizes. It was moved from the filter change to item change, because now the content not only changes with filter but when the user sets items

---

*[test/filtering.html, line 217 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-KUHbrcWVMjF7HxfAMt5:-KUXAFNFZHEOH2Zys08U:bbtaput) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/02f90f2098ea9d6358b36d255a5a956f1c59996f/test/filtering.html#L217)):*
> Exactly.
</details>
will add a test if the PR is relevant to merge. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUX8NeAeA15csapK0gA:b-bhp0d9)*

Saulis:
Had a closer look at the observers we have in place: when `items` is set with an array, no filtering is done do it. Filtering only happens after the `filter` changes. 

So, in practice, observing `filter` for changes -> passing it to an iron-ajax -> setting new `items` from the response -> by default behaves like there wouldn't be any internal filtering at all. 

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUYhJM46QhOJ8BWmHMc:b-g1iejn)*

manolo:
Between `filter for changes -> passing it to an iron-ajax ` There is an undesired filtering that modifies the content of the displayed list until the response arrives. 

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUYltH9091VyhoskOkt:bx5dr3r)*

Saulis:
Right, thats true. Need to look into this a bit more.

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/327#-:-KUa1_NXCDhQiR-4S4C3:b-2ojnb6)*

Saulis:
Continued in #350

PR #341, opened 2016-10-31
fixes #287: adds support for falsey primitive values when using Objec…

### LINKED ISSUE #287: 0 (Zero) values are not supported
if you have an items array like this `[{value: 0, label: 'pls support me'}, {value: 1, label: 'I am supported'}]` the `vaadin-combo-box` is bugging around when you want to select the item with the value 0.

I know there shouldn't be Entries in the Database with the ID 0, but it's an existing one so I can't change it.

Simple fix for this:

in the vaadin-combo-box-behavior.html on line 483 replace this function with this Code:

```
    _indexOfValue: function (value, items) {
          items = items || this.items;
          if (items) {
            for (var i = 0; i < items.length; i++) {
              if (String(this._getItemValue(items[i])) === String(value)) {
                return i;
              }
           }
         }

       return -1;
    },
```

this will also support null values (if you want that)

I can't do a pull request because i already have another change on my fork.

samiheikki:
…ts as Items

Primitives 0, -0 and false are now acceptable as values for the <vaadin-combo-box>. Example:

```
items = [ { value: 0, label: "zero" }, { value: false: "false" } ];
```

fixes #287: adds support for falsey primitive values - review fixes

- addresses the 6 comments in the pull request review
- adds explicit support for empty string values
- ensures that values like 0 and '0' are considered different (i.e. a number and a string representation of the same number)

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/341)

### DIFF
--- a/docs/vaadin-combo-box-objects.adoc
+++ b/docs/vaadin-combo-box-objects.adoc
@@ -20,6 +20,14 @@ var combobox = document.querySelector('vaadin-combo-box');
 combobox.items = [{ label: 'Display Value', value: 'Actual Value' }];
 ----
 
+The [propertyname]#value# property of the Object can hold String, Number or a Boolean primitive values. All truthy values as well as 0, -0 and false are supported. All other falsey values are not supported (such as NaN, undefined, null).
+
+[source,javascript]
+----
+var combobox = document.querySelector('vaadin-combo-box');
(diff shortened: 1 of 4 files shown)


Saulis:
Closed with 1cc8d72

PR #350, opened 2016-11-25
External Filtering

### LINKED ISSUE #259: Custom filtering

### LINKED ISSUE #260: Lazy loading of items
A database might have hundreds of thousands of rows of items that should be displayed in the combo box. Rather than always sending all of the items to the client, it should be possible to load the items lazily.

The first "page" of items could be loaded when the user opens the dropdown (focuses the field or clicks the toggle button). After that, new items/pages would be loaded on demand when the users either scrolls the or filters the list.

There should be a visual indication to the end user when new items are being loaded (perhaps a spinner in place of the toggle button).

### LINKED ISSUE #284: Expose input-element-value to facilitate AJAX requests

Saulis:
Fixes #284 
Fixes #259 
Fixes #260

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/350)


Saulis:
Question: should we also prevent keyboard navigation and tapping while combo-box is in "loading" state?

manolo:
Review status: 0 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-combo-box-behavior.html, line 35 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe0HP0W1f5d82sA-BZ:-KXe0HP0W1f5d82sA-B_:b7mzfum) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L35)):*
> ```HTML
>        * If set to `true`, when the `filter` changes, `<vaadin-combo-box>` will
>        * remain in loading state until a new, pre-filtered array of items is
>        * assigned to `items`.
> ```

Is it needed to assign a new array? wouldn't be enough if the user just modifies the content, or might be the content is not modified at all because the filter does not filters more elements ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Review status: 0 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXe2cBT5UkIAlsZLiqm:b-kphs0t) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*
> ```HTML
> <dom-module id="vaadin-spinner">
> ```

I fill like having our own component for spinning here is too much. Shouldn't we have a simpler solution like an icon or just an SVG like in split layout which could be replaceable by the user.  ping @platosha 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

tomivirkki:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 467 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeDrH3MQHSRprcOzV-:-KXeDrH3MQHSRprcOzV0:bgowp4q) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L467)):*
> ```HTML
>           var item = valueIndex > -1 && this.items[valueIndex];
>           if (item) {
>             //TODO: this is duplicate from _valueChanged
> ```

Hmm...were you meant to fix this TODO before submitting?

---

*[vaadin-combo-box-behavior.html, line 473 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeEOL5TnKATqwWr71Q:-KXeEOL5TnKATqwWr71R:b-vus8op) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L473)):*
> ```HTML
>           }
>         } else {
>           // this needs to be set async in case a external filter-changed
> ```

Shouldn't be needed. The "Custom Filtering" does just that and works fine (because lines 393 and 394 are in correct order).

---

*[vaadin-combo-box-overlay.html, line 181 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe-7Og7QfktLal4Rzs:-KXe-7Og7QfktLal4Rzt:b-7huaez) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-overlay.html#L181)):*
> ```HTML
> 
>     _onTap: function(e) {
>       this.fire('selection-changed', {item: this._items[e.model.index]});
> ```

Why not `{item: e.mode.item}`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

tomivirkki:
Review status: 0 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 12 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLA3mPWtLvFvbSLKh:-KXeLA3mPWtLvFvbSLKi:b-4podcs) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L12)):*
> ```AsciiDoc
> 
> By default, [vaadinelement]#vaadin-combo-box# is used with a local array of items
> is filtered internally to display only the items that match the input that the
> ```

Should the sentence on previous line continue here, in that case needs "and" at the beginning.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 24 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLWfMBS-3h9j9poF6:-KXeLWfNkkyxwqEPufUh:b1jr31x) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L24)):*
> ```AsciiDoc
> on the input field. To prevent that, we need to set [propertyname]#externalFiltering#.
> 
> This not only disable the client side filtering, but also tell
> ```

"disable" -> "disables", "tell" -> "tells"

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 46 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLrgchiVkIuG9TldI:-KXeLrgchiVkIuG9TldJ:b-x7lioy) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L46)):*
> ```AsciiDoc
> == Custom Filtering
> 
> When you want to custom filtering on the client side, you can listen for changes
> ```

"custom" -> "customize"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
Review status: 0 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 35 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe0HP0W1f5d82sA-BZ:-KXeM0nQ4mm7LXk4N88a:bswcdqn) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L35)):*

Is it needed to assign a new array? wouldn't be enough if the user just modifies the content, or might be the content is not modified at all because the filter does not filters more elements ?
</blockquote></details>

We need something that notifies the combo-box that the external filtering is now complete. In addition to reassigning the array, it actually does also work if you mutate the array using Polymer mutation functions, but I'm not sure if we want to promote that...?

---

*[vaadin-combo-box-behavior.html, line 467 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeDrH3MQHSRprcOzV-:-KXeTw-DDQaBDJBJjnfo:b-10fiv4) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L467)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Hmm...were you meant to fix this TODO before submitting?
</blockquote></details>

Not really.. at the time of writing the two block were more closely implemented than now. Also, don't know how to name the reusable parts.. gonna remove the comment.

---

*[vaadin-combo-box-behavior.html, line 473 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeEOL5TnKATqwWr71Q:-KXeNQlfX27hTtHZq2qJ:b-co4bs4) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L473)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Shouldn't be needed. The "Custom Filtering" does just that and works fine (because lines 393 and 394 are in correct order).
</blockquote></details>

You're absolutely right. Removed `async`.

---

*[vaadin-combo-box-overlay.html, line 181 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe-7Og7QfktLal4Rzs:-KXeS_7rGtP1oPebalZN:blo7lg3) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-overlay.html#L181)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Why not `{item: e.mode.item}`?
</blockquote></details>

I did have problems using `e.model.item` earlier – I'm now kicking myself for not writing the reasons down in the comment. Anyways, looks working just fine now with `e.model.item`

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXeURxOil3nTNctXH4h:b-oyhi6g) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*

I fill like having our own component for spinning here is too much. Shouldn't we have a simpler solution like an icon or just an SVG like in split layout which could be replaceable by the user.  ping @platosha 
</blockquote></details>

I wouldn't consider customising (apart from the colour) the loading spinner before there's an actual request for it. 

These styles were directly copied over from `vaadin-grid` 2.0, and I just wanted to wrap it in a custom element so it can be easily reused. I'm fine with changing the approach to something else as long as we can reuse it easily – however, unless there are any ready made solutions that could replace this spinner, I wouldn't spent any time on it at this point.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 12 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLA3mPWtLvFvbSLKh:-KXeVe6_sdHsi9Xc0k2U:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L12)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Should the sentence on previous line continue here, in that case needs "and" at the beginning.
</blockquote></details>

Done.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 24 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLWfMBS-3h9j9poF6:-KXeVmdRaHDyJoN0pAh5:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L24)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

"disable" -> "disables", "tell" -> "tells"
</blockquote></details>

Done.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 46 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeLrgchiVkIuG9TldI:-KXeVqHObfn38wn4xYP_:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/docs/vaadin-combo-box-external-filtering.adoc#L46)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

"custom" -> "customize"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Review status: 0 of 15 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXe_SDrTuo-Qyek8kJt:bpmubo2) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

I wouldn't consider customising (apart from the colour) the loading spinner before there's an actual request for it. 

These styles were directly copied over from `vaadin-grid` 2.0, and I just wanted to wrap it in a custom element so it can be easily reused. I'm fine with changing the approach to something else as long as we can reuse it easily – however, unless there are any ready made solutions that could replace this spinner, I wouldn't spent any time on it at this point.
</blockquote></details>

IMO, at least we should allow the user replace this by paper-spinner so as they can align with MD (https://material.google.com/components/progress-activity.html#progress-activity-types-of-indicators)

Hence, I suggest two solutions:

1. replace this with paper-spinner
2. enclose this in a `content select=` and leave the user change it with any implementation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXf4euTFQe3xdTFMnYq:b-9uhsab) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*

IMO, at least we should allow the user replace this by paper-spinner so as they can align with MD (https://material.google.com/components/progress-activity.html#progress-activity-types-of-indicators)

Hence, I suggest two solutions:

1. replace this with paper-spinner
2. enclose this in a `content select=` and leave the user change it with any implementation


</blockquote></details>

At this point, I don't think there is any nice way then extracting this into a separate component:

- ready-made `<paper-spinner>` providers a lot more features then we need here (nice cooling down, ARIA support).
- SVG animations are not widely supported and depricated in Chrome already, so CSS animations is still the best option so far. Since we’re doing CSS, it is nice to scope the styles with the DOM context in a web component, regardless of what we use to draw it, would it be SVG or just DOM.

But since we make it with reusability in mind, it would be nice to align the spinner’s API with `<paper-spinner>` now, so that they are more exchangeable within each other.

---

*[vaadin-spinner.html, line 18 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeNQ1lJ4_4yGuZ43yr:-KXeNQ1lJ4_4yGuZ43ys:b5ikhu6) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L18)):*
> ```HTML
>         border-top-color: transparent;
>         content: "";
>         height: 16px;
> ```

Btw, this feels way too small when compared with the other Polymer elements.

`<iron-icon>`’s default size is 24×24.
`<paper-spinner>` is 32×32.

---

*[vaadin-spinner.html, line 40 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXf6rSYdD5RqbMoDSwg:-KXf6rSZUM1xvvjm_tiw:b-seyp2r) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L40)):*
> ```HTML
> 
>       properties: {
>         loading: {
> ```

For example, this property could be changed to `active` to be more aligned with `<paper-spinner>` API.

Also, it’s not the appropriate name IMO, there is nothing that is loading in the spinner.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXf9KSBvXW-7YxxfKcN:b-gqxdhv) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*
> then
do you mean  s/then/than/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
Review status: 0 of 17 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-spinner.html, line 18 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeNQ1lJ4_4yGuZ43yr:-KXgZ2TE96SQC6F6WB7m:b-f1scm) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L18)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Btw, this feels way too small when compared with the other Polymer elements.

`<iron-icon>`’s default size is 24×24.
`<paper-spinner>` is 32×32.
</blockquote></details>

I've resized the overlay height to be `58px` while loading, the `24px` sized spinner seems to fit in nicely – please try it out.

---

*[vaadin-spinner.html, line 40 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXf6rSYdD5RqbMoDSwg:-KXgZ0JsJA7SUpZgi2Qr:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L40)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

For example, this property could be changed to `active` to be more aligned with `<paper-spinner>` API.

Also, it’s not the appropriate name IMO, there is nothing that is loading in the spinner.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
I feel like we could re-use the same idea than in the `iron-ajax`. They have a `loading` property which could be bound to a boolean status in the combo-box replacing the external filtering. In this way the user can use the spinner whenever he wants, and also we don't leave the component in an indeterminate status if the items array does not change,

---

Review status: 0 of 17 files reviewed at latest revision, 14 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 35 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe0HP0W1f5d82sA-BZ:-KXfWWonw0OCu5Q3cRh9:bwn835) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L35)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

We need something that notifies the combo-box that the external filtering is now complete. In addition to reassigning the array, it actually does also work if you mutate the array using Polymer mutation functions, but I'm not sure if we want to promote that...?
</blockquote></details>

I see this way confusing, I feel more in the way of having a property for setting the loading status.

---

*[demo/external-filtering.html, line 39 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfL_pugopmNNLxnL2a:-KXfL_pugopmNNLxnL2b:b-5fo637) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L39)):*
> ```HTML
>         document.addEventListener('WebComponentsReady', function() {
>           Polymer({
>             is: 'iron-ajax',
> ```

I do prefer a real iron-ajax example here so as the user can copy and paste, and we don't have to maintain this mock code. 
Otherwise you must remove iron-ajax dependency from bower

---

*[demo/external-filtering.html, line 74 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfMJvIEFdhcp0K-Ulk:-KXfMJvIEFdhcp0K-Ull:b-mmyub4) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L74)):*
> ```HTML
>       <template>
>         <template is="dom-bind">
>           <iron-ajax url="http://your.remote.url/?filter=[[filter]]" last-response={{response}} debounce-duration="500" auto>
> ```

although default is json, I would add the `handle-as=json` here

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 11 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfNe8L7Cckesbnd2m3:-KXfNe8L7Cckesbnd2m4:b-yi62c4) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L11)):*
> ```AsciiDoc
> == Overview
> 
> By default, [vaadinelement]#vaadin-combo-box# is used with a local array of items
> ```

I would remove the word local, because it could be remote, for instance if the list is gotten initially using iron-ajax, but the user doesn't need custom filters. For instance list of countries coming from a remote server but is enough to filter locally. 

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 13 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfOgv6a0spBR7rbBrf:-KXfOgv6a0spBR7rbBrg:bs53cjr) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L13)):*
> ```AsciiDoc
> By default, [vaadinelement]#vaadin-combo-box# is used with a local array of items
> that are filtered internally to display only the items that match the input the
> user has typed. The search algorithm used is a simple case insensitive search.
> ```

redundant search word, maybe replace with: `The default algorithm used is just a contains case insensitive search.`

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 21 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfQjseCy5oSNjO3sea:-KXfQjseCy5oSNjO3seb:blp3ucx) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L21)):*
> ```AsciiDoc
> == Remote Filtering
> 
> By default, the internal client side filtering is run every time user types something
> ```

every time the user ...

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 22 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfQxlkKUkBB3nX-ItZ:-KXfQxlkKUkBB3nX-It_:b5xyys0) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L22)):*
> ```AsciiDoc
> 
> By default, the internal client side filtering is run every time user types something
> on the input field. To prevent that, we need to set [propertyname]#externalFiltering#.
> ```

on -> in

... to set the externalFiltering property.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 26 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfRCrSN9fToxNiMLM0:-KXfRCrSN9fToxNiMLM1:b5lmomm) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L26)):*
> ```AsciiDoc
> This not only disables the client side filtering, but also tells
> [vaadinelement]#vaadin-combo-box# to remain in "loading" state until a fresh set
> of pre-filtered items is reassigned to [propertyname]#items#.
> ```

.. to the items property

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 28 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfRIGlbGeRODt8D2Pq:-KXfRIGlbGeRODt8D2Pr:bjzaw84) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L28)):*
> ```AsciiDoc
> of pre-filtered items is reassigned to [propertyname]#items#.
> 
> You can combine these properties together with [vaadinelement]#iron-ajax# as follows:
> ```

You can combine both properties together with the `iron-ajax` component as follows

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 46 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfRdtrfwJyxOpGVs7F:-KXfRdtrfwJyxOpGVs7G:b3d6y7g) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L46)):*
> ```AsciiDoc
> == Custom Filtering
> 
> When you want to customize filtering on the client side, you can listen for changes
> ```

when -> if 
listen for changes on -> listen to changes in 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-:-KXilhSr7EAIXspyLDSG:b9qwho9)*

manolo:
Review status: 0 of 17 files reviewed at latest revision, 16 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 393 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXiqcvQJZ-_IZHEyxQ7:-KXiqcvQJZ-_IZHEyxQ8:bl8ov26) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/vaadin-combo-box-behavior.html#L393)):*
> ```HTML
>           this._filterChanged(this.filter);
>         } else {
>           this._loading = this.externalFiltering;
> ```

A query with an empty value is sent before the combo box is opened, it should be sent when actually the user opens it. Otherwise all combo boxes in your UI will fire ajax requests when the page loads. This is not a lazy pattern. 

---

*[vaadin-combo-box-behavior.html, line 459 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXirNILjp7vevw3nWNj:-KXirNILjp7vevw3nWNk:brgfoo3) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/vaadin-combo-box-behavior.html#L459)):*
> ```HTML
>     _itemsChanged: function(e) {
>       if (e.path === 'items' || e.path === 'items.splices') {
>         this._setItems(this.items);
> ```

Setting items when the overlay is opened the first time does not update the visible items viewport.

Try your example and enclose the filter in a timeout of 2 seconds. Open the overlay before the first request with an empty value comes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Review status: 0 of 17 files reviewed at latest revision, 17 unresolved discussions.

---

*[demo/external-filtering.html, line 75 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXj0lm7VDy1KVegV7oG:-KXj0lm7VDy1KVegV7oH:b-4ck5ip) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L75)):*
> ```HTML
>         <template is="dom-bind">
>           <iron-ajax url="http://your.remote.url/?filter=[[filter]]" last-response={{response}} debounce-duration="500" auto>
>           </iron-ajax>
> ```

The implementation seems working wrong with real iron-ajax requests, see this http://g.recordit.co/lL1kyJvzAq.gif

- The first request is sent before the combobox is used
- The first actual filter does not have spinner at all
- The first response only shows 3 items, you need to close-open the overlay to update
- Zero item array should show an empty overlay, but shows nothing that looks confusing for the user. Is not something new but noticeable when data is async.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Review status: 0 of 17 files reviewed at latest revision, 17 unresolved discussions.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXjip3RRliGs1tR54UF:b-7kpc07) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*

> then
do you mean  s/then/than/
</blockquote></details>

sorry, yes,`s/then we need/than we need/`

---

*[vaadin-spinner.html, line 18 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeNQ1lJ4_4yGuZ43yr:-KXjrww4TRVX8_YlHMTf:bpqphbk) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L18)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

I've resized the overlay height to be `58px` while loading, the `24px` sized spinner seems to fit in nicely – please try it out.
</blockquote></details>

As discussed in person, `24px` fits nicely indeed, the only issue is that currently it’s not `24px` yet.

---

*[demo/external-filtering.html, line 74 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXjlZuzfli85gHctmeE:-KXjlZuzfli85gHctmeF:byqa4q6) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/bf5b835b1f414984854d5702b28bc97d62dce32f/demo/external-filtering.html#L74)):*
> ```HTML
>       <template>
>         <template is="dom-bind">
>           <iron-ajax url="http://your.remote.url/?filter=[[filter]]" last-response={{response}} debounce-duration="500" auto>
> ```

Please use the `example.com`, `example.net` or `example.org`. See rfc2606.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
That's a brilliant idea, I'll see if I can change the `externalFiltering` to `loading`

---

Review status: 0 of 17 files reviewed at latest revision, 17 unresolved discussions.

---

*[vaadin-spinner.html, line 18 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXeNQ1lJ4_4yGuZ43yr:-KXkM51SJL_TCerIRWSa:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L18)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

As discussed in person, `24px` fits nicely indeed, the only issue is that currently it’s not `24px` yet.
</blockquote></details>

Done.

---

*[demo/external-filtering.html, line 74 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfMJvIEFdhcp0K-Ulk:-KXkM7qCSgTCQuoyuVRM:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L74)):*

although default is json, I would add the `handle-as=json` here
</blockquote></details>

Done.

---

*[demo/external-filtering.html, line 75 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXj0lm7VDy1KVegV7oG:-KXkM9A_wMB-Hoe0TbLM:b-gep72e) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L75)):*

The implementation seems working wrong with real iron-ajax requests, see this http://g.recordit.co/lL1kyJvzAq.gif

- The first request is sent before the combobox is used
- The first actual filter does not have spinner at all
- The first response only shows 3 items, you need to close-open the overlay to update
- Zero item array should show an empty overlay, but shows nothing that looks confusing for the user. Is not something new but noticeable when data is async.
</blockquote></details>

Can you try if it works any better now?

---

*[demo/external-filtering.html, line 74 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXjlZuzfli85gHctmeE:-KXkM8h9v9waheyfV3Gk:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/bf5b835b1f414984854d5702b28bc97d62dce32f/demo/external-filtering.html#L74)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Please use the `example.com`, `example.net` or `example.org`. See rfc2606.
</blockquote></details>

Done.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 11 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfNe8L7Cckesbnd2m3:-KXkMFLEEUxQAYhpkmEX:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L11)):*

I would remove the word local, because it could be remote, for instance if the list is gotten initially using iron-ajax, but the user doesn't need custom filters. For instance list of countries coming from a remote server but is enough to filter locally. 
</blockquote></details>

Done.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 13 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfOgv6a0spBR7rbBrf:-KXkMGyMu5wnX80iNa7w:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L13)):*

redundant search word, maybe replace with: `The default algorithm used is just a contains case insensitive search.`
</blockquote></details>

Done.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 21 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfQjseCy5oSNjO3sea:-KXkMHSO1nWdho8nrpXf:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/docs/vaadin-combo-box-external-filtering.adoc#L21)):*

every time the user ...
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-:-KXj1_R6A8R8DRlOqWtB:b-do87a1)*

platosha:
Reviewed 5 of 16 files at r2, 3 of 6 files at r4, 8 of 8 files at r5.
Review status: 16 of 17 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

*[vaadin-combo-box-behavior.html, line 14 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXk_idfptuOEFjd2uRv:-KXk_idfptuOEFjd2uRw:b-nw4jql) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L14)):*
> ```HTML
>     properties: {
>       /**
>        * An array of values to be filtered for as options in the dropdown. The
> ```

Basically, I’d like to make it more clear, when to use `items` or `displayItems`.

Maybe rephrase this to put less emphasis on ”options in the dropdown”? Maybe mention that this is a source set or a full set of items? For ex: “A full set of items to filter the options from. The items can be of either...”

---

*[vaadin-combo-box-behavior.html, line 33 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkZoEPFxNFR-mULVIU:-KXkZoEQCai_jMB8oN5r:b-656rtj) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L33)):*
> ```HTML
> 
>       /**
>        * An array of values to be displayed in the list of options. The
> ```

Should we differentiate this more from `items`, like: “A subset of items filtered based on the user input. Displayed in the list of options.”?

---

*[vaadin-combo-box-behavior.html, line 68 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkbwTT1nEpu-ejt40Y:-KXkbwTT1nEpu-ejt40Z:b6esi1c) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L68)):*
> ```HTML
> 
>       /*
>        * When set to `true`, as loading spinner is displayed on top of the list of options.
> ```

typo: “as loading” → “a loading”

---

*[vaadin-combo-box-behavior.html, line 81 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkc9iPhbBvWnqpEQGn:-KXkc9iPhbBvWnqpEQGo:b-z52y80) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L81)):*
> ```HTML
> 
>       /**
>        * String used for filtering items.
> ```

I would also mention that this comes from the user (from the input element) somehow.

---

*[vaadin-combo-box-behavior.html, line 558 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkf13TJ76LJVjSH5Qg:-KXkf13TJ76LJVjSH5Qh:bo1816j) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L558)):*
> ```HTML
>       var selectedItem = detail.item;
>       if (selectedItem !== null) {
>         this.value = this._getItemValue(selectedItem);
> ```

So, within these changes, the workflow is: 
- user selects an item in the overlay
- we extract and set the `value` for the item
- the `value` observer tries to find the `selectedItem` for that `value`

Previously, the `selectedItem` worked as a source of truth, and now its the string `value`. I don’t think this is a good idea.

First, it is redundant to search for the `selectedItem` after we already have it here. 

Second, what would happen if we have more then one item with the same string `value`?

Since the `selectedItem` is the source object, and the `value` is straightforwardly extracted from it, isn’t it more natural to keep the state in the `selectedItem`, but not in it’s outcome `value`?

---

*[vaadin-combo-box-overlay.html, line 74 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkhNJD_8ksjdgaZNyq:-KXkhNJD_8ksjdgaZNyr:bn9frik) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/5bbce95f2dbcd0306190932d5543b63dc54dd421/vaadin-combo-box-overlay.html#L74)):*
> ```HTML
>           selected-item="{{_selectedItem}}"
>           items="[[_items]]"
>           selection-enabled
> ```

Why not use `<iron-list>` selection anymore?

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 22 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkkFBrC98Gziuwm7A9:-KXkkFBrC98Gziuwm7AA:b6izol1) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/docs/vaadin-combo-box-external-filtering.adoc#L22)):*
> ```AsciiDoc
> 
> By default, the internal client side filtering is run every time the user types something
> into the input field. To bypass that, you can set an pre-filtered array of items into [propertyname]#filteredItems#.
> ```

Should we also note somewhere in this doc, that one don’t need to set `items` when using `filteredItems`?

---

*[test/toggling-dropdown.html, line 219 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXknlxbVedMeZNjO5WU:-KXknlxcew93zbAhYJiy:b-cfgka8) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/toggling-dropdown.html#L219)):*
> ```HTML
>       }
> 
>       if (touchDevice) {
> ```

The tap is fired on desktop as well.

So as the click event is fired on touch devices.

These two tests should probably be run on all the devices.

---

*[test/using-object-values.html, line 61 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkoOh2ZNG5DNTFQpfx:-KXkoOh2ZNG5DNTFQpfy:b-ywgo7m) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/using-object-values.html#L61)):*
> ```HTML
>       function selectItem(index) {
>         // simulates tapping on the overlay items, but it more reliable in tests.
>         combobox.$.overlay.fire('selection-changed', {item: combobox.items[index]});
> ```

Hey, I can’t see where we test that the `selection-change` event is fired.

---

*[test/using-object-values.html, line 130 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkpHYQ-SZcSxmgeosU:-KXkpHYR_rrJjGUEMxZ4:brcyazb) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/using-object-values.html#L130)):*
> ```HTML
> 
>       it('should set the selected item', function() {
>         combobox.value = 'bar';
> ```

This should work on the closed combo-box as well, but now we have `.open` in the `beforeEach` block.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
Review status: 12 of 17 files reviewed at latest revision, 22 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 393 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXiqcvQJZ-_IZHEyxQ7:-KXoYedC_AQ88WWchSpN:b-t21076) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/vaadin-combo-box-behavior.html#L393)):*

A query with an empty value is sent before the combo box is opened, it should be sent when actually the user opens it. Otherwise all combo boxes in your UI will fire ajax requests when the page loads. This is not a lazy pattern. 
</blockquote></details>

It's not your typical lazy pattern, but it leads to better UX since the overlay will open faster. If needed, developer can listen to `opened` property being set to `true` before fetching any items. 

---

*[vaadin-combo-box-behavior.html, line 459 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXirNILjp7vevw3nWNj:-KXoZn7MQyhOZejs37Us:bt2vr1b) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/vaadin-combo-box-behavior.html#L459)):*

Setting items when the overlay is opened the first time does not update the visible items viewport.

Try your example and enclose the filter in a timeout of 2 seconds. Open the overlay before the first request with an empty value comes.
</blockquote></details>

Fixed.

---

*[vaadin-combo-box-behavior.html, line 14 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXk_idfptuOEFjd2uRv:-KXoVFp3gOVp_qHVHmQ_:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L14)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Basically, I’d like to make it more clear, when to use `items` or `displayItems`.

Maybe rephrase this to put less emphasis on ”options in the dropdown”? Maybe mention that this is a source set or a full set of items? For ex: “A full set of items to filter the options from. The items can be of either...”
</blockquote></details>

Done.

---

*[vaadin-combo-box-behavior.html, line 33 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkZoEPFxNFR-mULVIU:-KXoXwz0KaYCaAswFaCy:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L33)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Should we differentiate this more from `items`, like: “A subset of items filtered based on the user input. Displayed in the list of options.”?
</blockquote></details>

Done.

---

*[vaadin-combo-box-behavior.html, line 68 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkbwTT1nEpu-ejt40Y:-KXoY-KcdpfowhadB43T:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L68)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

typo: “as loading” → “a loading”
</blockquote></details>

Done.

---

*[vaadin-combo-box-behavior.html, line 81 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkc9iPhbBvWnqpEQGn:-KXoYSRsLymNQH3Z3ypM:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L81)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

I would also mention that this comes from the user (from the input element) somehow.
</blockquote></details>

Done.

---

*[vaadin-combo-box-behavior.html, line 558 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkf13TJ76LJVjSH5Qg:-KXo_4qBjoUH6ryD9n0o:bu8bxe) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L558)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

So, within these changes, the workflow is: 
- user selects an item in the overlay
- we extract and set the `value` for the item
- the `value` observer tries to find the `selectedItem` for that `value`

Previously, the `selectedItem` worked as a source of truth, and now its the string `value`. I don’t think this is a good idea.

First, it is redundant to search for the `selectedItem` after we already have it here. 

Second, what would happen if we have more then one item with the same string `value`?

Since the `selectedItem` is the source object, and the `value` is straightforwardly extracted from it, isn’t it more natural to keep the state in the `selectedItem`, but not in it’s outcome `value`?
</blockquote></details>

Currently `selectedItem` is `readOnly` and `value` is the property which user uses to define the selection programmatically. In the previous implementation, this caused the selection data to flow two ways: from `value` to `selectedItem` and the selection in the overlay, at the same time user clicks would flow the selection in the opposite direction.

I agree that it would make more sense having `selectedItem` as the source of truth, but that would require us to change the `readOnly` properties of these two – hence I would do that separately from this PR.

---

*[vaadin-combo-box-overlay.html, line 74 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkhNJD_8ksjdgaZNyq:-KXoarHtqqRH-hCrfeBk:b-hfwe11) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/5bbce95f2dbcd0306190932d5543b63dc54dd421/vaadin-combo-box-overlay.html#L74)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Why not use `<iron-list>` selection anymore?
</blockquote></details>

We should store the selection state in the combo-box instead – using `iron-list`'s selection behaviour is the cause for example having those `listen`/`unlisten` calls for `selected-item-changed` event inside the `setItems()`, which I've now removed. 

---

*[demo/external-filtering.html, line 39 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXfL_pugopmNNLxnL2a:-KXobZzZs-IDT-Lg-drk:b-z02bar) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L39)):*

I do prefer a real iron-ajax example here so as the user can copy and paste, and we don't have to maintain this mock code. 
Otherwise you must remove iron-ajax dependency from bower
</blockquote></details>

I'll remove it. I was bumping into throttling issues when using some real remote backend. Also, it's more coherent to use the "Elements" data as we do with every other demo.

---

*[docs/vaadin-combo-box-external-filtering.adoc, line 22 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkkFBrC98Gziuwm7A9:-KXocLenqgllvFosmoMV:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/docs/vaadin-combo-box-external-filtering.adoc#L22)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Should we also note somewhere in this doc, that one don’t need to set `items` when using `filteredItems`?
</blockquote></details>

Done.

---

*[test/using-object-values.html, line 61 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkoOh2ZNG5DNTFQpfx:-KXogAFvLPNZBFyPQ9Uw:bqmjwcm) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/using-object-values.html#L61)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Hey, I can’t see where we test that the `selection-change` event is fired.
</blockquote></details>

Added!

---

*[test/using-object-values.html, line 130 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkpHYQ-SZcSxmgeosU:-KXogk-r1-r5y8qinPyk:b-pupvje) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/using-object-values.html#L130)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

This should work on the closed combo-box as well, but now we have `.open` in the `beforeEach` block.
</blockquote></details>

Added a separate test for closed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Reviewed 5 of 5 files at r6.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 350 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXopRs_PRwpuowMiYZv:-KXopRsato4vb8nZistJ:b-wdziqw) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/313a73674663a229cc50a02cf3262dd19df3f7f3/vaadin-combo-box-behavior.html#L350)):*
> ```HTML
> 
>     _onOpened: function() {
>       this.$.overlay.hidden = !this._hasItems(this.$.overlay._items) && !this.loading;
> ```

So, effectively, we hide the overlay until the items are loaded? Why?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 350 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXopRs_PRwpuowMiYZv:-KXpOVRTQZ7_-ZDXQs8W:bz385wu) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/313a73674663a229cc50a02cf3262dd19df3f7f3/vaadin-combo-box-behavior.html#L350)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

So, effectively, we hide the overlay until the items are loaded? Why?
</blockquote></details>

OK, understood.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 558 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXkf13TJ76LJVjSH5Qg:-KXpSeCQPwzkWxod5Zsg:b-gc42y0) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/vaadin-combo-box-behavior.html#L558)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

Currently `selectedItem` is `readOnly` and `value` is the property which user uses to define the selection programmatically. In the previous implementation, this caused the selection data to flow two ways: from `value` to `selectedItem` and the selection in the overlay, at the same time user clicks would flow the selection in the opposite direction.

I agree that it would make more sense having `selectedItem` as the source of truth, but that would require us to change the `readOnly` properties of these two – hence I would do that separately from this PR.
</blockquote></details>

OK to fix in a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
Reviewed 2 of 2 files at r8.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Reviewed 4 of 16 files at r2, 1 of 3 files at r3, 3 of 6 files at r4, 5 of 8 files at r5, 3 of 5 files at r6, 2 of 2 files at r8.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 35 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe0HP0W1f5d82sA-BZ:-KXpYkCz2TNB7JkIohJN:b7yqxza) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-combo-box-behavior.html#L35)):*

I see this way confusing, I feel more in the way of having a property for setting the loading status.
</blockquote></details>

solved with the filtered-items prop

---

*[vaadin-combo-box-behavior.html, line 393 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXiqcvQJZ-_IZHEyxQ7:-KXpZF5Fg-3Ov1ogI_7O:b-cihjmo) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/vaadin-combo-box-behavior.html#L393)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

It's not your typical lazy pattern, but it leads to better UX since the overlay will open faster. If needed, developer can listen to `opened` property being set to `true` before fetching any items. 
</blockquote></details>

opened does not happen when filter is fired, so the user has to filter with the overlay closed.
But anyway user can handle the empty value somehow in code.

---

*[vaadin-combo-box-behavior.html, line 350 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXopRs_PRwpuowMiYZv:-KXoqY25KSESjwOC9fHU:b-sdjmr9) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/313a73674663a229cc50a02cf3262dd19df3f7f3/vaadin-combo-box-behavior.html#L350)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

OK, understood.
</blockquote></details>

user can show the overlay setting loading to true, if he wants some feedback.

---

*[vaadin-spinner.html, line 1 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXe2cBT5UkIAlsZLiql:-KXorSQr1sIVvNmWOPm6:b2vfz5p) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/ed2f17c05bc2da8f2297aff9f247875ae4178247/vaadin-spinner.html#L1)):*
> they are more exchangeable within each other

Then, shouldn't we have a mechanism to this

---

*[demo/external-filtering.html, line 75 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXj0lm7VDy1KVegV7oG:-KXpTYNVDnxcYIETS0hi:b-93bacg) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/e352296ef7a034dd5609ff469283f09cfb57b560/demo/external-filtering.html#L75)):*
<details><summary><i>Previously, Saulis (Sauli Tähkäpää) wrote…</i></summary><blockquote>

Can you try if it works any better now?
</blockquote></details>

1- can be handled by user
2,4 - can now be managed via the loading prop
3- still remains, but as commented in stalker, it is already in master, so no blocker for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
Review status: 16 of 17 files reviewed at latest revision, 5 unresolved discussions.

---

*[test/toggling-dropdown.html, line 219 at r5](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXknlxbVedMeZNjO5WU:-KXu-oh3JuulJtAIKYLP:b-896fix) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/181285d6897e9fae806eb542f7ddade3e56bc137/test/toggling-dropdown.html#L219)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

The tap is fired on desktop as well.

So as the click event is fired on touch devices.

These two tests should probably be run on all the devices.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

manolo:
Reviewed 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

tomivirkki:
*[vaadin-combo-box-behavior.html, line 432 at r9](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXuU7LU2Y6Ekzxldk0h:-KXuU7LU2Y6Ekzxldk0i:b-l6b1bi) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/08138a97d9743cfc2e574d27739381b00b2c7b0f/vaadin-combo-box-behavior.html#L432)):*
> ```HTML
> 
>     _loadingChanged: function(loading) {
>       this.$.overlay.$.scroller.hidden = loading;
> ```

Overlay already has a bound and reflected `loading` property. Why not bind the `loading` further to the scroller's `hidden` declaratively or hide it in the overlay's CSS?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

Saulis:
Review status: 15 of 17 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 432 at r9](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-KXuU7LU2Y6Ekzxldk0h:-KXudv45xVpC17_wxcm-:bxorf74) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/08138a97d9743cfc2e574d27739381b00b2c7b0f/vaadin-combo-box-behavior.html#L432)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Overlay already has a bound and reflected `loading` property. Why not bind the `loading` further to the scroller's `hidden` declaratively or hide it in the overlay's CSS?
</blockquote></details>

Excellent idea, moved also `notifyResize` call and making overlay visible there!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

tomivirkki:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-:-KXy_VM2WyHQilxLvZ7-:bnfp4nl)*

manolo:
Reviewed 2 of 2 files at r10.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350)*

platosha:
:lgtm:

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/350#-:-KXydChocl9uqihMovYk:bnfp4nl)*

PR #402, opened 2017-02-14
Expose mixins on vaadin-combo-box-overlay and vaadin-combo-box-item 

ellenaua:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/402)

### DIFF
--- a/vaadin-combo-box-overlay.html
+++ b/vaadin-combo-box-overlay.html
@@ -27,6 +27,8 @@
         */
         z-index: 200;
         overflow: hidden;
+
+        @apply(--vaadin-combo-box-overlay-mixin);
       }
 
       #scroller {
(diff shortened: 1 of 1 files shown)


tomivirkki:
Hi @ellenaua 

Thanks for the PR. Exposing style mixins for `<vaadin-combo-box>` overlay and it’s contents is a bit problematic since the overlay is physically detached from `<vaadin-combo-box>` itself (it’s moved under `<body>` on open).

So once the overlay gets moved the mixin values defined for `<vaadin-combo-box>` no longer affect the overlay. Because of how the polyfill is currently implemented, the mixins seem to work but if you run `Polymer.updateStyles()` after opening the overlay, the scopes get updated and the problem occurs.

So adding mixins to the overlay would mean exposing a feature that relies on a certain polyfill implementation which isn’t very safe.

jouni:
Since this PR hasn’t made any progress in a few months, I’m closing it for now. Add comments if you feel we should take another look.

PR #416, opened 2017-03-13
Make `change` event to fire only on close and clear

### LINKED ISSUE #385: change event fired without user interaction
According to MDN documents

`The change event is fired for <input>, <select>, and <textarea> elements when a change to the element's value is committed by the user. Unlike the input event, the change event is not necessarily fired for each change to an element's value.`

The vaadin-combo-box implementation isn't following this contract as has been advertised. 

### Expected outcome
The event should only fire when value changes as result of a user gesture 

### Actual outcome
The event is fired when value changes programattically

platosha:
Fixes #385

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/416)

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -13,6 +13,7 @@
   <link rel="import" href="common.html">
 
   <link rel="import" href="../../iron-form/iron-form.html">
+  <link rel="import" href="../../paper-button/paper-button.html">
 </head>
 
 <body unresolved>
@@ -66,6 +67,30 @@
           form.addEventListener('iron-form-submit', function() {
             alert('Form submitted with name: ' + form.serialize().name);
             return false;
+          });
+        </script>
+      </template>
+    </demo-snippet>
+
+    <h3>Change Event</h3>
+    <demo-snippet>
+      <template>
+        <vaadin-combo-box id="changeEventDemo" items="[[elements]]"></vaadin-combo-box>
+        <paper-button id="setRandomValue" raised>Set random value</paper-button>
+        <script>
+          var combobox4 = document.querySelector('#changeEventDemo');
+          combobox4.items = elements;
+
+          combobox4.addEventListener('change', function(e) {
+            if (combobox4.value) {
+              alert('User selected value ' + combobox4.value);
+            } else {
+              alert('User has cleared the value');
+            }
+          });
+
+          document.querySelector('#setRandomValue').addEventListener('click', function() {
+            combobox4.value = combobox4.items[Math.floor(Math.random() * combobox4.items.length)];
           });
         </script>
       </template>
--- a/test/selecting-items.html
+++ b/test/selecting-items.html
@@ -194,32 +194,82 @@
       expect(valueChangedSpy.callCount).to.equal(1);
     });
 
-    it('should fire exactly one `change` event', function() {
-      var spy = sinon.spy();
-      combobox.addEventListener('change', spy);
+    describe('`change` event', function() {
+      it('should fire `value-changed` and `selected-item-changed` before `changed`', function(done) {
+        combobox.addEventListener('change', function() {
+          expect(valueChangedSpy.called).to.be.true;
+          expect(selectedItemChangedSpy.called).to.be.true;
+          done();
+        });
 
-      combobox.value = 'foo';
-
-      expect(spy.callCount).to.equal(1);
-    });
-
-    it('should fire `value-changed` and `selected-item-changed` before `changed`', function(done) {
-      combobox.addEventListener('change', function() {
-        expect(valueChangedSpy.called).to.be.true;
-        expect(selectedItemChangedSpy.called).to.be.true;
-        done();
+        combobox.open();
+        combobox.value = 'foo';
+        combobox.close();
       });
 
-      combobox.value = 'foo';
-    });
+      it('should fire exactly one `change` event', function() {
+        var spy = sinon.spy();
+        combobox.addEventListener('change', spy);
 
-    it('should stop input `change` event from bubbling', function() {
-      var spy = sinon.spy();
-      combobox.addEventListener('change', spy);
+        combobox.open();
+        combobox.value = 'foo';
+        combobox.close();
 
-      combobox.$.input.fire('change');
+        expect(spy.callCount).to.equal(1);
+      });
 
-      expect(spy.callCount).to.equal(0);
+      it('should fire on clear', function() {
+        var spy = sinon.spy();
+        combobox.addEventListener('change', spy);
+
+        combobox.value = 'foo';
+        combobox.$.clearIcon.click();
+
+        expect(spy.callCount).to.equal(1);
+      });
+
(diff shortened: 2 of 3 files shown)


manolo:

---

*[vaadin-combo-box-behavior.html, line 412 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-KfCYukzq4tYcqgvMeGq:-KfCYukzq4tYcqgvMeGr:b23vsux) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/500e2056c602c829eba73bfca22ddf03d43e4784/vaadin-combo-box-behavior.html#L412)):*
> ```HTML
>       }
> 
>       this._dispatchChange();
> ```

Not sure how the code should be, but at first view I feel like dispatchChange only should be fired from methods that actually know whether the value was changed, calling always in _onClosed makes not sense to me. Just for thinking

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

platosha:
Reviewed 1 of 1 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-combo-box-behavior.html, line 412 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-KfCYukzq4tYcqgvMeGq:-KfH0fUsbk7GTajPSXrR:b-g14i5k) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/500e2056c602c829eba73bfca22ddf03d43e4784/vaadin-combo-box-behavior.html#L412)):*

Not sure how the code should be, but at first view I feel like dispatchChange only should be fired from methods that actually know whether the value was changed, calling always in _onClosed makes not sense to me. Just for thinking
</blockquote></details>

Maybe misnamed, but `_dispatchChange()` contains additional change detection. In case of cancelling the dropdown with <kbd>Esc</kbd>, the value does not change, `_dispatchChange()` detects that and does not dispatch.

Calling on close here helps to detect user action in more safe way then dispatching on `_valueChanged`, for example.

Renamed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

tomivirkki:
*[vaadin-combo-box-behavior.html, line 527 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-Kff4-VS-A6-jrKxZ1Pn:-Kff4-VS-A6-jrKxZ1Po:bbr0l8u) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/08ddceceee8b45677302152d1ee25d2ff6d3763a/vaadin-combo-box-behavior.html#L527)):*
> ```HTML
>       }
>       // In the next _detectAndDispatchChange() call, the change detection should pass
>       this._lastCommittedValue = undefined;
> ```

Closing after the value is changed programmatically will always fire `change`. For example the following test won't pass:
```
it('should not fire after value changed programmatically', function() {
  combobox.value = 'foo';

  var spy = sinon.spy();
  combobox.addEventListener('change', spy);

  combobox.open();
  combobox.close();

  expect(spy.callCount).to.equal(0);
});
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

platosha:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-combo-box-behavior.html, line 527 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-Kff4-VS-A6-jrKxZ1Pn:-Kff5PBCJUHk4vVcQ2XW:bz8vota) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/08ddceceee8b45677302152d1ee25d2ff6d3763a/vaadin-combo-box-behavior.html#L527)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary><blockquote>

Closing after the value is changed programmatically will always fire `change`. For example the following test won't pass:
```
it('should not fire after value changed programmatically', function() {
  combobox.value = 'foo';

  var spy = sinon.spy();
  combobox.addEventListener('change', spy);

  combobox.open();
  combobox.close();

  expect(spy.callCount).to.equal(0);
});
```
</blockquote></details>

Intentional. We can not easily detect if the value change was done programmatically or by user.

Within this PR, when the dropdown closes, the value is assumed as committed by user, even if the value itself actually was set programmatically. I. e., the assumption is that close means user commits the value.

This will, obviously, dispatches false change if **close** is done programmatically. I suggest that we just accept this as a limitation for now, as non-user-introduced closing programmatically combined together with critical reliance on proper `change` is an edge case of edge cases. Let us assume that dropdown is always closed by user. A proper detection of whether closing is done indeed by user or programmatically would be complex once again, on par with user change detection itself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

platosha:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-combo-box-behavior.html, line 527 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-Kff4-VS-A6-jrKxZ1Pn:-KffJxU15cR4WbakiMuL:b3o586n) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/08ddceceee8b45677302152d1ee25d2ff6d3763a/vaadin-combo-box-behavior.html#L527)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Intentional. We can not easily detect if the value change was done programmatically or by user.

Within this PR, when the dropdown closes, the value is assumed as committed by user, even if the value itself actually was set programmatically. I. e., the assumption is that close means user commits the value.

This will, obviously, dispatches false change if **close** is done programmatically. I suggest that we just accept this as a limitation for now, as non-user-introduced closing programmatically combined together with critical reliance on proper `change` is an edge case of edge cases. Let us assume that dropdown is always closed by user. A proper detection of whether closing is done indeed by user or programmatically would be complex once again, on par with user change detection itself.
</blockquote></details>

Changed as discussed. Now it considers only changes after opening, and does not dispatch if dropdown was cancelled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

tomivirkki:
:lgtm:

---

Reviewed 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416#-:-KffMcTBaO9Wxbbo7uY1:bnfp4nl)*

manolo:
Reviewed 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/416)*

PR #418, opened 2017-03-14
Fix hidden items in overlay

JosefJezek:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/418)

### DIFF
--- a/vaadin-combo-box-overlay.html
+++ b/vaadin-combo-box-overlay.html
@@ -29,6 +29,10 @@
         overflow: hidden;
       }
 
+      [hidden] {
+        display: none !important;
+      }
+
       #scroller {
         overflow: auto;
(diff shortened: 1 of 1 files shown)


tomivirkki:
Hi @JosefJezek, thanks for the PR. For me however the hidden items are properly hidden without an additional style. Is there some specific case or browser where you’re having issues with the hidden items?

JosefJezek:
Yes, when use custom filter using `filtered-items` property. :-/

tomivirkki:
At least the "Custom Filtering" demo at https://vaadin.com/elements/-/element/vaadin-combo-box#demos uses the `filtered-items` property and properly hides non-visible items.

Can you provide a test case with your PR that validates the fix (a test that breaks before and passes after the change)?

jouni:
Closing because there hasn’t been any activity in almost 2 months.

PR #423, opened 2017-03-20
Disable dropdown focused item

JosefJezek:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/423)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vaadin-combo-box",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "description": "Polymer element for displaying a list of items with filtering",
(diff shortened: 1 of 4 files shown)


tomivirkki:
Thanks for the pull request. Can you elaborate a bit on the change, what it does and what's it's use-case?

With a quick look, it seems to add a new flag to the combo-box API that, when switched on, disables the input field from previewing the value while the user navigates the items with arrow keys. Did I understand correctly?

Saulis:
Is this use case something that can be achieved using the item templates?

jouni:
Closing because there hasn’t been any activity in almost 2 months.

Issue #424, opened 2017-03-21
Feature request: add option to hide clear icon


web-padawan:
### Description
In some cases we expect to always have some value selected, e. g. user should not be allowed to clear, to emulate the same appearance as [paper-dropdown-menu](https://www.webcomponents.org/element/PolymerElements/paper-dropdown-menu) actually has.

I know that we can use `vaadin-combo-box-light` for that, but since the expected change is small, what about implementing it as a boolean property, e. g. 
```js
noClearIcon: {
  type: Boolean,
  reflectToAttribute: true
}
````

It should be as simple as adding a few CSS rules: 
```css
:host([no-clear-icon]) paper-input-container paper-icon-button.clear-button,
:host([no-clear-icon]) paper-input-container ::content paper-icon-button.clear-button {
  display: none;
}

:host([has-value][no-clear-icon]) #input {
  padding-right: 32px;
}
```
What do you think? Is this worth implementing? If so, I could submit a PR.

tomivirkki:
Could this functionality be incorporated with the current `required` property so that when set to `true`, the field wouldn’t be clearable? WDYT @jouni?

jouni:
Initially sounds like a nice idea. But I’m a bit wary of coupling features together too much, as it doesn’t leave these decisions up for the user/developer if there are varying use cases that should be supported.

My suggestion would be to expose a CSS mixin for the clear button instead, which would allow the developer to hide the button using CSS.

Even though we are planning on deprecating all CSS mixins in an upcoming minor, it would be okay for Polymer 1 users. And we are still going to offer the same possibility, but with the proprietary theming solution we are planning to add in the next major version.

web-padawan:
@jouni good point. But hiding buttons means also need to decrease padding, as noted above.
Adding another mixin to achieve that second goal looks excessive and not very DRY.
So, which solution shall we end up with for a moment?

jouni:
Oh, right – I didn’t realize we are using padding on the input to compensate for the buttons. What’s the reason for doing that, @tomivirkki? Shouldn’t paper-input-container do that for us, or aren’t we using the suffix element feature for the buttons?

jouni:
> Adding another mixin to achieve that second goal looks excessive and not very DRY.
What second goal are you referring to? Sorry, I’m a bit lost :)

jouni:
This is of course an obvious workaround for now, although it doesn’t take care of the unnecessary padding on the input:

```html
[...]

PR #419, opened 2017-03-14
use slots instead of content

manolo:
Connected to #413

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/419)

PR #425, opened 2017-03-22
Use suffix slot for clear and toggle icons

web-padawan:
Just as @jouni has mentioned [here](https://github.com/vaadin/vaadin-combo-box/issues/424#issuecomment-288422095).

Please check if I removed correct styles, especially those related to clear icon (there used to be some different padding and negative `bottom`).

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/425)

### DIFF
--- a/vaadin-combo-box-shared-styles.html
+++ b/vaadin-combo-box-shared-styles.html
@@ -20,7 +20,7 @@
 
       paper-icon-button.small,
       :host ::content paper-icon-button.small {
-        padding: 6px 6px !important;
+        padding: 6px !important;
       }
 
       :host(:not([has-value])) ::content [slot=clear-button],
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -98,10 +98,6 @@ Custom property | Description | Default
       paper-icon-button.toggle-button,
       :host ::content paper-icon-button.clear-button,
       :host ::content paper-icon-button.toggle-button {
-        position: absolute;
-        bottom: -4px;
-        right: -4px;
-
         line-height: 18px !important;
         width: 32px;
         height: 32px;
@@ -112,11 +108,6 @@ Custom property | Description | Default
         cursor: pointer;
         margin-top: -1px;
         --paper-icon-button-ink-color: rgba(0, 0, 0, .54);
-      }
-
-      paper-input-container paper-icon-button.clear-button,
-      paper-input-container ::content paper-icon-button.clear-button {
-        right: 28px;
       }
 
       paper-input-container paper-icon-button:hover,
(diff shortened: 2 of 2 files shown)


tomivirkki:
:lgtm: Great update, date picker probably needs the same thing

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/425#-:-KfuyZBKedJS9Egyld2k:b-ua6zoc)*

web-padawan:
this PR will conflict with #419. /cc @manolo 
Should I wait until work on replacing `content` with `slots` gets merged and then rebase?

web-padawan:
FYI: I'm going to rebase this, shall I wait until new release or better do it right now?

Saulis:
@web-padawan I think we should do a rebase, merge and release this PR now – the Polymer 2 support will take some time still before it's done.

web-padawan:
Rebased.

limonte:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/425#-:-KgYsBg_H4JDNex92Usb:bnfp4nl)*

tomivirkki:
Ok, I just remembered why we positioned the icons the way they were before. Too bad we didn't have a test to enforce that :(
Here's an example where you can see the issue (additional prefix stretches the component): http://jsbin.com/bupoqifize/edit?html,output

jouni:
Darn, completely forgot that also! Can we compensate that somehow else, for example by having a fixed default width for the element that matches the previous default width?

web-padawan:
I should consider that use case when submitting this PR too :(
What if we restore absolute positioning, but use another approach for input sizing? E. g.
```css
:host([has-value]) #input {
  width: calc(100% - 64px)
}
```
This would still require a workaroung to hide clear icon, like passing hidden div and overriding styles like this:
```css
--paper-input-container-input: {
  width: calc(100% - 32px) !important;
}
```
At least this looks more transparent than `padding` tricks.

jouni:
I wonder how many users would be affected by the fact that the element width changes when you select a value, if you’ve set the width to `auto`? It would not exactly be breaking any promises, but it is of course a change in behavior. I also wonder what would be the use case to use `width: auto`.

jouni:
What I’m trying to say is that I would like to keep this suffix-based implementation instead of absolute positioning if somehow possible.

web-padawan:
@jouni  I haven't found any suggestions in MD guidelines on this topic. But in general, using form elements as inline-block is not very common, and `paper-input` does nothing about width when inserting suffix.

jouni:
I’m okay with changing the behavior. It requires a minor version at least, and some sort of beta period to gather feedback if other users have issues with it.

PR #429, opened 2017-03-28
Allow specifying the width for the overlay.

panuhorsmalahti:
Sometimes it's very useful to specify the overlay dropdown width.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/429)

### DIFF
--- a/vaadin-combo-box-behavior.html
+++ b/vaadin-combo-box-behavior.html
@@ -139,6 +139,13 @@
         type: HTMLElement,
         readOnly: true
       },
+
+      /**
+       * The overlay width in pixels. If not used, the width of the dropdown is used.
+       */
+      overlayWidth: {
+        type: Number
+      },
       _toggleElement: Object,
       _clearElement: Object,
 
--- a/vaadin-combo-box-light.html
+++ b/vaadin-combo-box-light.html
@@ -66,7 +66,8 @@ two `<paper-button>`s to act as the clear and toggle controls.
         loading="[[loading]]"
         on-down="_onOverlayDown"
         on-mousedown="_preventDefault"
-        vertical-offset="[[overlayVerticalOffset]]">
+        vertical-offset="[[overlayVerticalOffset]]"
+        width="[[overlayWidth]]">
     </vaadin-combo-box-overlay>
   </template>
 </dom-module>
--- a/vaadin-combo-box.html
+++ b/vaadin-combo-box.html
@@ -228,7 +228,8 @@ Custom property | Description | Default
         on-down="_onOverlayDown"
         loading="[[loading]]"
         on-mousedown="_preventDefault"
-        vertical-offset="2">
+        vertical-offset="2"
+        width="[[overlayWidth]]">
     </vaadin-combo-box-overlay>
   </template>
 </dom-module>
--- a/vaadin-overlay-behavior.html
+++ b/vaadin-overlay-behavior.html
@@ -16,6 +16,13 @@
       },
 
       /**
+       * Width of the overlay, if not set the width of positionTarget is used
+       */
+      width: {
+        type: Number
+      },
+
+      /**
        * Vertical offset for the overlay position.
        */
       verticalOffset: {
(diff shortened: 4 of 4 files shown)


jouni:
I agree, control over the width behavior is sometimes needed. But adding a pixel based property for it feels wrong.

I would feel better if this control was in CSS, maybe with a custom property targeting the overlay, like `--vaadin-combo-box-overlay-width`. By default that would be `input-width` (or something similar) and the current computed value would be used (matching the width of the input). Then you could set that to any CSS value and that would be used instead, or you could set it to `auto` to make it size itself with the content/items inside the overlay (but still have a min-width that matches the input element).

panuhorsmalahti:
Yeah, I could use `getComputedStyleValue("--vaadin-combo-box-overlay-width")` instead of `this.width`.

jouni:
> Yeah, I could use getComputedStyleValue("--vaadin-combo-box-overlay-width") instead of this.width.

That would be a step in the right direction, but I would like to avoid the inline style completely. After considering this a bit more, here’s what I would want us to do:

- Instead of setting the `width` of the overlay to `positionTarget.clientWidth`, set it to `minWidth` instead. I think the overlay should never be smaller than the input element itself. If you disagree, can you provide a use case where this would be beneficial?
- Add a new CSS custom property in `vaadin-combo-box-overlay.html` to the `:host` element, `width: var(--vaadin-combo-box-overlay-width);`

Saulis:
There's a scoping issue with using a CSS property since the overlay gets placed under `<body>`. 
It might work with an initial value in Shady DOM but not in Shadow I'm afraid.

jouni:
Crap, always forget this scoping issue. I guess there’s no bullet-proof way to workaround that – like copying certain CSS properties from the host element to the overlay element when the overlay is opened?

Saulis:
Well it could work, but it's a bit hairy 💇 

What about an approach where we add a property which makes the overlay to take its width from the item width instead of the input width?

jouni:
@Saulis, too hairy to even consider? Or would you be willing to do it, biting teeth? 😬 

I mean, the issue isn’t limited to this particular styling aspect 🙁 

If we want to support some of the upcoming theming properties, such as `--theme-font-size` or `--theme-grid-size`, so that they would seem to be inherited by the overlay as well even when set on the host element, we need some solution to those as well. And my current understanding is that we might not even know the names of those properties in advance.

jouni:
That might be an overkill, and we should just support the proposed `theme-class` attribute, and copy that to the overlay and then let a custom “theme module” specify the required CSS property values for the overlay.

To recap, I still feel we should always have the min-width of the overlay match the input width. We can have a custom property to control that, like `--vaadin-combo-box-overlay-min-width`, which would not be set by default (`getComputedStyle` returns undefined).

Whether the overlay should be allowed to have a fixed width, or let itself grow with the items, could be controlled with a CSS custom property like `--vaadin-combo-box-overlay-width`. 

The default value would be empty (`getComputedStyle` would return undefined) which means the width is limited to the input width. If set to `auto`, the overlay would be allowed to grow with the content. It would also be allowed to be set to any valid CSS length to have a fixed width.

jouni:
Well, after doing a bit of prototyping, `width: auto;` doesn’t really work after all, since the overlay items inside iron-list are `position: absolute` and don’t grow the container. Also, when scrolling, longer items would grow the overlay during scrolling even if they would not be absolutely positioned. So the items need to be measured if that’s something we want to support.

jouni:
You can take a look at the proto and comment on that. I added some initial documentation also, but I didn’t add any tests for it yet.

Saulis:
@panuhorsmalahti did you try if Jouni's prototype would work for you:

https://github.com/vaadin/vaadin-combo-box/commit/816d2447ae44d8cfac32a29c4212c60edc613773

panuhorsmalahti:
The prototype looks suitable for me.

Saulis:
Replaced this PR with #444

PR #437, opened 2017-04-28
Fix overlay scrolling via the scrollbar in IE

kisstkondoros:
In IE clicking on the scrollbar causes the internal input element to loose the focus and close the overlay.
This is unintended, and with this additional check it can be avoided.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/437)

### DIFF
--- a/vaadin-combo-box-behavior.html
+++ b/vaadin-combo-box-behavior.html
@@ -179,7 +179,7 @@
     },
 
     _onBlur: function() {
-      if (!this._closeOnBlurIsPrevented && document.activeElement !== this.$.overlay.$.scroller) {
+      if (!this._closeOnBlurIsPrevented && document.activeElement !== this.$.overlay.$.scroller && document.activeElement !== null) {
         this.close();
       }
     },
--- a/vaadin-combo-box-behavior.html
+++ b/vaadin-combo-box-behavior.html
@@ -179,7 +179,7 @@
     },
 
     _onBlur: function() {
-      if (!this._closeOnBlurIsPrevented) {
+      if (!this._closeOnBlurIsPrevented && document.activeElement !== this.$.overlay.$.scroller) {
         this.close();
       }
     },


manolo:
What version of IE has the issue? I checked in Edge-15 and couldn't reproduce.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/437#-:-KkZViJ2ij2T-CvAlWoN:bjgns36)*

kisstkondoros:
It can be easily reproduced using IE 11 once you get the whole thing working :)
e.g. https://cdn.vaadin.com/vaadin-core-elements/master/vaadin-combo-box/demo/index.html
does not work in IE 11.

Some additional changes are needed, I'll adjust it in a minute.

manolo:
Actually, in that URL IE11 seems not working, I cannot open the menu at all by clicking or typing in the input, can you? or it's an issue in my version?

kisstkondoros:
No it isn't working in IE11 at all, I'm not sure what changes are necessary to get it working. I'll check at work today.

kisstkondoros:
After a bit of investigation it seems that the linked demo set shadow dom explicitly, therefore it does not work in IE11, adding ?dom=shady to the url reveals another issue with the setup, which I haven't investigated further.

But!
There is another demo site linked at [vaadin-combobox#demos](https://vaadin.com/elements/-/element/vaadin-combo-box#demos) which is the following:
[jsbin demo](https://jsbin.com/licubadeko/1/edit?html,output)

Here the issue can be reproduced without any further setup.

kisstkondoros:
Closed accidentally 😄

kisstkondoros:
The problem with the [vaadin-combobox-demo](https://cdn.vaadin.com/vaadin-core-elements/master/vaadin-combo-box/demo/index.html) seems to be unrelated to shadow dom. It is not forced to use shadow dom, it was a false assumption.

platosha:
Hi @kisstkondoros, thanks for you contribution! Your commits were merged.

I have elaborated on top of your code a bit, mostly for easier maintenance reasons. Here is what have changed:

1. I’ve noticed that the `document.activeElement` is different with Polymer 2.0 and 1.x, thus requiring checking different cases. Luckily, I found another solution with the `unselectable` attribute, which works nicely and identical with different Polymer versions. This way we don’t have to revisit this place when dropping Polymer 1.x support in the next major release.

2. Described the fixed issue in the comment, so that we remember why are these lines there in future.

3. Added a test to prevent possible reverts and regressions.

4. The fix and the test are scoped for IE only, as this does not happen in other browsers.

Issue #344, opened 2016-11-09
Combo Box not working in IE 11.0.36


jay8t6:
The dropdown works, but none of the items are selectable on the latest version of IE. Can you guys take a look?

Browser: IE 
Version: 11.0.36

samiheikki:
For me it is working in this example: http://plnkr.co/edit/TFtYLzNQQYvu2lQUH0PO?p=preview. Can you provide an example where the problem occurs?

jay8t6:
We tried your demos also, its broken. The latest IE11 patch might have something to do with it. Make sure you test it out on 11.0.36

samiheikki:
I filed a separated issue on our demos, see #345. But was the issue in the plunker example that I attached in the previous message. If yes, can you provide a screenshot?

jay8t6:
@samiheikki I am having my team check on this. I will let you know shortly. Thanks for putting together the plunkr.

jay8t6:
@samiheikki Sorry about the delay. but, we were able to confirm that this happens when testing on a Windows 8.1 machine running IE 11 (Version 11.0.9600.18161). Hope that helps.

limonte:
I also can confirm the issue in Win8/IE10:

![recorded](https://cloud.githubusercontent.com/assets/6059356/21217438/e5a4df54-c2b4-11e6-9197-b7f520dd6228.gif)

BTW, it's weird that the same example works fine in Win10/IE11

jay8t6:
@samiheikki any updates on this issue?

m4po:
Hi guys!
we use the combo-box element in our project and we have also the problem to select the items in the overlay section in IE11 (Win8). I have fixed my problem by overwriting the __setPosition method in vaadin-combo-box-overlay-behavior (v1.3.3).
In line 136, translate3d is used, which does not work properly in IE11. 

My solution:
This.style.top = this._translateY + 'px';
This.style.left = this._translateX + 'px';

instead of
This.translate3d (this._translateX + 'px', this._translateY + 'px', '0');

Because the overlay is positioned absolute or fixed!
Works also fine in Chrome, FF and Edge. I think translate3d isn't necessary!

tomivirkki:
Works with the latest version. Closing.

PR #441, opened 2017-05-03
Change CSS-positioning of overlay

samuelportz:
As in #344  mentioned when opening the vaadin-combo-box-overlay in IE 11 the translate3d isn't working as expected. The items are not selectable.
When using the top and left-values it is working as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/441)

### DIFF
--- a/vaadin-overlay-behavior.html
+++ b/vaadin-overlay-behavior.html
@@ -133,7 +133,8 @@
       var _devicePixelRatio = window.devicePixelRatio || 1;
       this._translateX = Math.round(this._translateX * _devicePixelRatio) / _devicePixelRatio;
       this._translateY = Math.round(this._translateY * _devicePixelRatio) / _devicePixelRatio;
-      this.translate3d(this._translateX + 'px', this._translateY + 'px', '0');
+      this.style.top = this._translateY + 'px';
+      this.style.left = this._translateX + 'px';
 
       this.style.width = this.positionTarget.clientWidth + 'px';
 


manolo:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-overlay-behavior.html, line 136 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/441#-KkZTN1iVqkxeQ8EEB6A:-KkZTN1iVqkxeQ8EEB6B:bgmxpyj) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/29620067d654999f778616f2821b63ca82023ea9/vaadin-overlay-behavior.html#L136)):*
> ```HTML
>       this._translateX = Math.round(this._translateX * _devicePixelRatio) / _devicePixelRatio;
>       this._translateY = Math.round(this._translateY * _devicePixelRatio) / _devicePixelRatio;
>       this.translate3d(this._translateX + 'px', this._translateY + 'px', '0');
> ```

Isn't translate3d supported in IE11?  actually, it seems it was introduced in IE10, (https://msdn.microsoft.com/en-us/library/jj200286(v=vs.85).aspx). 
Probably the problem here is that you need to change the last parameter to add the units, can you test if  changing `'0'` by `'0px'` works for you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/441)*

samuelportz:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[vaadin-overlay-behavior.html, line 136 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/441#-KkZTN1iVqkxeQ8EEB6A:-KkipgtIzC9H9DFyS06f:b-cd0fml) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/29620067d654999f778616f2821b63ca82023ea9/vaadin-overlay-behavior.html#L136)):*

Isn't translate3d supported in IE11?  actually, it seems it was introduced in IE10, (https://msdn.microsoft.com/en-us/library/jj200286(v=vs.85).aspx). 
Probably the problem here is that you need to change the last parameter to add the units, can you test if  changing `'0'` by `'0px'` works for you?
</blockquote></details>

It is supported but it seems not correctly implemented. Both `'0'` and `'0px'` have the same result. When the overlay is opened you see it at the expected position but the highlighted area is where you have to click to select an item. Clicking on the visible overlay will cause it to close (like an outside click).



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/441)*

jouni:
Ping @limonte, could you take a look at this and decide what to do, thanks!

tomivirkki:
Thank you for the contribution. I guess the rendering issue has been fixed for IE11 at some point as selecting items seems to work fine now (tried with 11.0.85/Win 10 and 11.0.7/Win 8.1) even with older versions of the combobox. Closing.

PR #455, opened 2017-05-27
Add filter property to item template

### LINKED ISSUE #430: Expose the filter property inside the item template
### Description
In this example https://gist.github.com/Saulis/48229751ba472e949b75eedc3d84de5b - say I include several `vaadin-combo-box` elements on the same page using the `regex-highlighter` component. The `filter="{{filter}}"` attribute and data binding is shared between the `vaadin-combo-box`, causing them all to rerun the regex match. In my use case, it is difficult to rename them as they are dynamically created.

@Saulis recommended https://github.com/vaadin/vaadin-combo-box/issues/106#issuecomment-289701115

> To enable your use case properly, we would need to expose the filter property inside the item template – which we probably will during the implementation of this feature

Saulis:
Fixes #430

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/455)

### DIFF
--- a/demo/item-template.html
+++ b/demo/item-template.html
@@ -40,14 +40,14 @@
       </template>
     </demo-snippet>
 
-    <h3>Focused and Selected Flags</h3>
+    <h3>Focused and Selected Flags and Filter</h3>
 
     <demo-snippet>
       <template>
         <vaadin-combo-box id="elements-selected-box" label="Element" value="Carbon">
           <template>
             <b>[[item]]</b><br>
-            selected: [[selected]], focused: [[focused]]
+            selected: [[selected]], focused: [[focused]], filter: [[filter]]
           </template>
         </vaadin-combo-box>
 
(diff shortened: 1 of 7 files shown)


platosha:
LGTM per se. Not sure if we want to add new features in the hybrid `<vaadin-combo-box>` though.

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[demo/item-template.html, line 43 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/455#-KlN_qNmenTV_Dbm0f5E:-KlN_qNnzHG1bDYJFk2Q:b18eauw) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/59f1ce5ac9ba41ca89cd12f7ee81aecee8cba8f6/demo/item-template.html#L43)):*
> ```HTML
>     </demo-snippet>
> 
>     <h3>Focused and Selected Flags and Filter</h3>
> ```

Should we maybe have a separate demo for the actual use case of the filter binding?

It does not quite fit right here, as `selected` and `focused` give different values for different items, while `filter` is the same for every item.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/455#-:-KlNbUbkAhLfCJvEjf2G:b1nujnx)*

PR #482, opened 2017-07-11
replace `vaadin.` with `window.vaadin.`

nkrishnaswami:
Using both `window.vaadin` and `vaadin` to refer to the same namespace breaks compiling in closure. This has been encountered and worked around/fixed in `vaadin-grid` by using a consistent name (`window.vaadin`) in https://github.com/vaadin/vaadin-grid/pull/872/

This patch applies the analogous change to `vaadin-combo-box`.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/482)

### DIFF
--- a/vaadin-combo-box-behavior.html
+++ b/vaadin-combo-box-behavior.html
@@ -4,11 +4,11 @@
 
 <script>
   window.vaadin = window.vaadin || {};
-  vaadin.elements = vaadin.elements || {};
-  vaadin.elements.combobox = vaadin.elements.combobox || {};
+  window.vaadin.elements = window.vaadin.elements || {};
+  window.vaadin.elements.combobox = window.vaadin.elements.combobox || {};
 
-  /** @polymerBehavior vaadin.elements.combobox.ComboBoxBehavior */
-  vaadin.elements.combobox.ComboBoxBehaviorImpl = {
+  /** @polymerBehavior window.vaadin.elements.combobox.ComboBoxBehavior */
+  window.vaadin.elements.combobox.ComboBoxBehaviorImpl = {
     properties: {
       /**
        * A full set of items to filter the visible options from.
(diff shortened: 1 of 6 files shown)


tomivirkki:
Related issue: https://github.com/vaadin/vaadin-grid/pull/886

Looks like replacing `vaadin.` with `window.vaadin.` alone didn't fix the issue in vaadin-grid after all.

nkrishnaswami:
Thanks, @limonte & @tomivirkki (also, very nice components)!
Tomi, do you mean the lint failures in the linked PR? 
Also, this patch works using closure, but I haven't (oops) confirmed that it's still ok without it; I will run a quick test of the demo page with `polymer serve` or similar to verify that, too, and comment again.

nkrishnaswami:
Lint & `polymer serve` look good; I'm requesting permission to sign the CLA now.

tomivirkki:
Hi @nkrishnaswami,

Didn’t mean lint, but overall Closure compilation.
So, if the intention is to stay compatible with Closure in future also (is this something we want to commit to @limonte? Should this be changed in the Skeleton and all other components as well?), a compilation step should probably be included in our build process (that's run automatically by the CI).

I created a draft implementation for the linked PR at https://github.com/vaadin/vaadin-grid/commit/ea965bba4eb468ca7533ea78c8099e912c00e788 a while back. Not 100% sure I got it right there, but it looks like the compilation doesn't pass without additional changes. So something similar should be included in this PR also in order to
1. avoid a regression breaking the compatibility in future commits 
2. verify that the changes in the PR fix the original problem

It's kind of the same deal as with normal bug fixes; we require a unit test alongside each bug fix PR for the same exact reasons.

nkrishnaswami:
Got it; I'll look at the draft implementation and add a closure test to the pull req.

nkrishnaswami:
Bad news -- work doesn't permit me to sign the CLA (copyright assignment), so I'll go ahead and close this.  If that changes, I'll make a new patch with the required base/testcase.

PR #485, opened 2017-07-18
Merge 2.0 branch into master

limonte:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/485)


manolo:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[vaadin-combo-box-item.html, line 91 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/485#-Kpn7s18tGRgRZzQ4rJi:-Kpn7s18tGRgRZzQ4rJj:b-b8gqs5) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/82ec9fa0046c033e47bb9bea2570def24680e8f1/vaadin-combo-box-item.html#L91)):*
> ```HTML
>       if (!this._itemTemplateInstance) {
>         // 2.0 has __dataHost
>         var comboBox = this.domHost.dataHost || this.domHost.__dataHost;
> ```

Probably we dont need to check old API here, so __dataHost  should be enough

---

*[test/form-input.html, line 98 at r4](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/485#-Kpn7kAZhU3ybCvA3Zj6:-Kpn7kAZhU3ybCvA3Zj7:bn9s140) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/82ec9fa0046c033e47bb9bea2570def24680e8f1/test/form-input.html#L98)):*
> ```HTML
>         comboBox.value = 'bar';
> 
>         var val = ironForm.serializeForm().foo;
> ```

inline this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/485)*

manolo:
After thinking a bit more, I don't understand why do we need this PR.

Basically this seems to port the fix for the FF problem duplicating the template content in the hybrid version to pure version 2. But this does not have the issue since it was already fixed. So in principle I would abandon this.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/485#-:-Kpoax956m8UbxwsOTm2:b-7ixncu)*

PR #496, opened 2017-07-28
Add functionality for disabling separated items

### LINKED ISSUE #443: Allow to disable items
I'm using `vaadin-combo-box` with `paper-item` as item template and I'd like to make some items disabled, so that they can't be selected. When I set the `disabled` attribute in a `paper-item` it displays like if it is disabled, but `vaadin-combo-box` still allows it to be selected.

Is there any way to make an item unselectable?

yuriy-fix:
Fixes #443

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/496)

### DIFF
--- a/vaadin-combo-box-item.html
+++ b/vaadin-combo-box-item.html
@@ -17,11 +17,11 @@ This program is available under Apache License Version 2.0, available at https:/
          display: none;
       }
 
-      :host([disabled]:hover){
+      :host([disabled]:hover) {
         background: none !important;
       }
 
-      :host([disabled]){
+      :host([disabled]) {
         color: #ccc;
       }
     </style>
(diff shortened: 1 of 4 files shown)


vaadin-combo-box-item.html
```diff
@@ -16,6 +16,14 @@
       :host([hidden]) {
          display: none;
       }
+
+      :host([disabled]:hover) {
+        background: none !important;
```
jouni: !important should be avoided. Use `:not([disabled])` instead for the original hver effect instead.

vaadin-combo-box-item.html
```diff
@@ -70,6 +78,15 @@
             reflectToAttribute: true
           },
 
+          /*
+           * True when item is disabled
+           */
+          disabled: {
+            type: Boolean,
+            value: false,
+            reflectToAttribute: true,
```
jouni: Remember to document this new state which as available for styling and part of the public API

tomivirkki:
Closing as inactive

PR #540, opened 2017-09-25
Disallow scrolling for vaadin-combo-box overlay

### LINKED ISSUE #537: Double scroll bar when an external mouse is connected
When I select a vaadin-combo-box and an external mouse is connected, two scrollbars appear instead of one (see attached screenshot).

This can also be confirmed on https://vaadin.com/elements/vaadin-combo-box/html-examples/combo-box-basic-demos

### Browsers Affected
Confirmed present in Chrome and Firefox. Not present in Safari.

screen shot 2017-09-22 at 6 03 46 pm

tomivirkki:
Fixes #537

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/540)

### DIFF
--- a/vaadin-combo-box-dropdown.html
+++ b/vaadin-combo-box-dropdown.html
@@ -25,6 +25,16 @@ This program is available under Apache License Version 2.0, available at https:/
   </template>
 </dom-module>
 
+<dom-module id="vaadin-combo-box-overlay-styles" theme-for="vaadin-combo-box-overlay">
+  <template>
+    <style>
+      [part="overlay"] {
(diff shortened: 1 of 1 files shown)


limonte:
Now `iron-list` is just cut from the bottom. To illustrate why we should not do this I increased the padding for `[part="content"]` to `20px`:

tomivirkki:
Maybe it's better to have this fix in the theme instead. Closing.

jouni:
Umm, I’m not sure how this should be fixed in the theme. Can you elaborate?

PR #555, opened 2017-10-20
Don't do anything in the initial _valueChanged observer

### LINKED ISSUE #554: Combobox doesn't show the selected item when the selectedItem property is set while the combobox is not appended to the DOM
### Description
If you set the property `selectedItem` before adding the combobox to the DOM, the property seems to be  reset - the component doesn't show the selected item.

This directly affects how the component works with Flow. See https://github.com/vaadin/flow/issues/2547

### Expected outcome
The component should show the selection no matter at what time the property was set (before or after attaching it to the DOM)

### Actual outcome
The input field inside the combobox remains empty if you set the property before attaching the component.

### Live Demo
https://jsfiddle.net/6r5htd05/64/

Click on `Append combo` and then click on `Set item` -> it works as expected.
Reset the demo
Then click on  `Set item` before and then `Append combo` -> the value is not shown on the combobox

(You can verify the selected item set by clicking on `Verify item`)

### Browsers Affected
Tested on the browsers below:
- [ x ] Chrome
- [ x ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

limonte:
Fixes #554

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/555)

### DIFF
--- a/test/selecting-items.html
+++ b/test/selecting-items.html
@@ -144,6 +144,14 @@
       combobox.close();
     });
 
+    it('should be possible to set selectedItem before attaching to the DOM', () => {
+      var newComboBox = combobox.cloneNode(true);
+      newComboBox.items = ['foo', 'bar'];
+      newComboBox.selectedItem = 'foo';
+      combobox.parentElement.appendChild(newComboBox);
+      expect(newComboBox.value).to.equal('foo');
+    });
+
     describe('`change` event', () => {
       it('should fire `value-changed` and `selected-item-changed` before `changed`', done => {
         combobox.addEventListener('change', () => {
--- a/vaadin-combo-box-mixin.html
+++ b/vaadin-combo-box-mixin.html
@@ -625,7 +625,11 @@ This program is available under Apache License Version 2.0, available at https:/
       this._focusedIndex = this.filteredItems.indexOf(selectedItem);
     }
 
-    _valueChanged(value) {
+    _valueChanged(value, oldVal) {
+      if (value === '' && oldVal === undefined) { // initializing, no need to do anything (#554)
+        return;
+      }
+
       if (this._isValidValue(value)) {
         let item;
         if (this._getItemValue(this.selectedItem) !== value) {


tomivirkki:
The fix looks good but needs a test case

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-KwtahuJQbBTot8jeyma:b-292uij)*

limonte:
I don't know how to write a test case for this change.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-KwtoKPxI8dR2GF6NPnp:bra49cx)*

manolo:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KwxhKtMLwe8-W3vFiwL:b-ptusfa) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
> ```HTML
>       if (value === '' && oldVal === undefined) { // initializing, no need to do anything (#554)
>         return;
>       }
> ```

It's confusing what this does, even with the comment.
What happens if you manually call `_valueChanged(this.value)` in the attach phase? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

gilberto-torrezan:
A test based on the bug that motivated this change would be enough? For instance:

1. Create the element using JS (`document.createElement('vaadin-combo-box')`)
2. Set some items inside
3. Set the `selectedItem`
4. Append the element to the DOM (you can append using timeout or on an event handler to make sure it is executed in a separate stage)
5. Assert that the `selectedItem` property is set after the element was appended, and that the `input` inside the combo-box contains the selected item text

I don't know how your actual testing works, so I'm just speculating.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-Kx7R13NHcB9StWP5Kub:b-637wt9)*

YuriyVaadin:
I suggest one of the possible ways will be to clone existing `comboBox` (as it will be cloned without `selectedItem`) and set the `selectedItem` before attaching it to the fixture. Something like this:
```
var newComboBox = comboBox.cloneNode(true);
newComboBox.selectedItem = 'foo';
comboBox.parentElement.replaceChild(newComboBox, comboBox);
```

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-KxC_Uvw3u3UMs9Xm3MH:bh2jygc)*

limonte:
Thanks for the idea @yuriyvaadin :beers: Done.

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxISG9TWfjSUahoTdWK:bwsrvuh) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
> What happens if you manually call _valueChanged(this.value) in the attach phase?

`_valueChanged()` is the observer method, not meant for manual calls.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-KxIS5sMqqOIMRSEJde7:b5frc9z)*

manolo:

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxNEFo3AsPYRxCRmVie:bl8fspv) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

> What happens if you manually call _valueChanged(this.value) in the attach phase?

`_valueChanged()` is the observer method, not meant for manual calls.
</blockquote></details>

Why not?, if is it a semantic issue, then the name can be adapted.

There is no convention about method in observers should have either a specific name, or to be only specific for that task.

Actually is a technique that we use in many places,  for instance `_updateAriaExpanded` in `ready` and `open` observer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

limonte:

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxNICZwjesLTj0Pzh-2:b-hm7tev) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
> Why not?, if is it a semantic issue, then the name can be adapted.

By definition. It's the private **observer** method. Calling it manually is the antipattern which should be avoided. If you want it triggered, change the property.

> Actually is a technique that we use in many places,  for instance _updateAriaExpanded in ready and open observer.

Link to source code or didn't happen.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

manolo:

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxNN-CE6mRO4iABf2Jf:b-gcc7b) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

> Why not?, if is it a semantic issue, then the name can be adapted.

By definition. It's the private **observer** method. Calling it manually is the antipattern which should be avoided. If you want it triggered, change the property.

> Actually is a technique that we use in many places,  for instance _updateAriaExpanded in ready and open observer.

Link to source code or didn't happen.
</blockquote></details>

- Where is the documentation defining that calling a method in ready as an antipattern ? 
- Main issue here is that the property setter and observer is called before ready because of the Polymer workflow, so we cannot do anything here. That the reason you need the hack
- https://github.com/vaadin/vaadin-combo-box/blob/master/vaadin-combo-box.html#L288

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

limonte:

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxNQ-GkFVBYye4Aig8h:b-1whm91) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
> Where is the documentation defining that calling a method in ready as an antipattern ?

Calling a method in ready is fine of course. Calling the **observer** method **manually** is antipattern, in ready or anywhere else. 

> Main issue here is that the property setter and observer is called before ready because of the Polymer workflow, so we cannot do anything here. That the reason you need the hack

Does the existence of a hack in one place mean that we should support the hack everywhere? I really hope that you don't think like that :) 

> https://github.com/vaadin/vaadin-combo-box/blob/master/vaadin-combo-box.html#L288

Great, hacks without even explanatory comments why it's there :/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

tomivirkki:
:lgtm:

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-:-KxNfguMwcDuUlH-aXef:bnfp4nl)*

manolo:

---

*[vaadin-combo-box-mixin.html, line 631 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555#-KwxhKtLFe9z-pdlo1M9:-KxS184DW6kiB0_nq6Kw:bfrei3i) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/b9bd7035d3c853d03a5d9e49fe615da01c94a916/vaadin-combo-box-mixin.html#L631)):*
<details><summary><i>Previously, limonte (Limon Monte) wrote…</i></summary><blockquote>

> Where is the documentation defining that calling a method in ready as an antipattern ?

Calling a method in ready is fine of course. Calling the **observer** method **manually** is antipattern, in ready or anywhere else. 

> Main issue here is that the property setter and observer is called before ready because of the Polymer workflow, so we cannot do anything here. That the reason you need the hack

Does the existence of a hack in one place mean that we should support the hack everywhere? I really hope that you don't think like that :) 

> https://github.com/vaadin/vaadin-combo-box/blob/master/vaadin-combo-box.html#L288

Great, hacks without even explanatory comments why it's there :/
</blockquote></details>

An observer is a method like is defined in polymer, don't understand why you want to make differences between class methods and class methods called from an observer declaration. Also I don't understand why you make undocumented assertions about what is a pattern or antipattern is in this matter-

There are plenty of cases where methods are shared in initialisation and observers, but seems useless for you this reason.

I just wanted to make this PR more readable, having to leave a comment with an issue reference to be able to understand what the code does is seems bad.
But seems that this discussion is unnecessary, so it's a matter of taste.

To me this code is not good :-1:  to go to master, but I don't want to be a blocker.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

manolo:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/555)*

PR #606, opened 2018-01-26
Set tabindex attribute on vaadin-combo-box-item

web-padawan:
The `vaadin-item` Lumo theme depends on `tabindex` attribute here: https://github.com/vaadin/vaadin-item/blob/master/theme/lumo/vaadin-item.html#L23

In the dropdown-menu the attribute is managed by list-mixin but we don't use it here, so we have to set this attribute on combo-box items in markup.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/606)


YuriyVaadin:
:lgtm: :thumbsup: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/606#-:-L3mcwzlLfzzsyMu9ylx:b-wpm7nb)*

jouni:
Ah, thanks for catching this. I already fixed this in vaadin-item styles, with `:host([role="option"])::before`, but this fix is more approriate IMO.

PR #619, opened 2018-02-22
Add content part with block context to the item

platosha:
Connected to vaadin/vaadin-item#17

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/619)

### DIFF
--- a/analysis.json
+++ b/analysis.json
@@ -30,11 +30,11 @@
               "sourceRange": {
                 "file": "../bower_components/vaadin-overlay/src/vaadin-overlay.html",
                 "start": {
-                  "line": 129,
+                  "line": 133,
                   "column": 10
                 },
                 "end": {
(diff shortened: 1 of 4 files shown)


jouni:
Is the `<span id="content">` element necessary? Would be nicer if simple string labels would not introduce one more element in the DOM.

platosha:
@jouni sadly we cannot use data binding for `[[label]]` in that case. But that’s fine, perhaps. Removed the span.

web-padawan:
:lgtm:

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/619#-:-L5xcslLQTb4NvqJflH2:bnfp4nl)*

PR #620, opened 2018-02-27
Set clear button not opening the overlay

### LINKED ISSUE #613: Don’t open the overlay when clearing the field using the “clear” button
### *Proposal for discussion*

The current behavior (focusing the field and opening the overlay) assumes the user would more often want to enter a new value when clearing the field.

A better assumption is that the user just wants to clear the field (common for filtering fields).

If they want to enter a new value, it’s more likely they will focus the field directly, and in that case, the overlay will open, so it’s quick to select another value from it.

The field should still become focused when pressing the “clear“ button with a mouse (current behavior). But, when pressing the “clear” button on a touch device, the field should not be focused to avoid the on-screen keyboard from popping up (which would be annoying when clearing filtering fields for example).

YuriyVaadin:
Fixes #613 

Not sure if I can properly test the focused state of the input on mobile and desktop. Will be grateful for any suggestions.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/620)

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -217,6 +217,15 @@ This program is available under Apache License Version 2.0, available at https:/
       this.addEventListener('click', this._onClick.bind(this));
 
       this.$.overlay.addEventListener('vaadin-overlay-touch-start', this._onOverlayTouchStart.bind(this));
+
+      this.addEventListener('touchend', e => {
+        if (this._clearElement && e.composedPath()[0] !== this._clearElement) {
+          return;
+        }
+
+        e.preventDefault();
+        this._clear();
+      });
     }
 
     /**
@@ -438,7 +447,6 @@ This program is available under Apache License Version 2.0, available at https:/
 
       if (!this.opened) {
         this._detectAndDispatchChange();
-        this.open();
       }
     }
 
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -280,6 +280,7 @@ This program is available under Apache License Version 2.0, available at https:/
       const path = e.composedPath();
       if (path.indexOf(this._clearElement) !== -1) {
         this._clear();
+        this.focus();
       } else if (path.indexOf(this.inputElement) !== -1) {
         if (path.indexOf(this._toggleElement) > -1 && this.opened) {
           this.close();
--- a/test/filtering.html
+++ b/test/filtering.html
@@ -49,7 +49,7 @@
         expect(comboBox.opened).to.equal(true);
       });
 
-      it('should open the popup when input field is cleared', () => {
+      it('should open the popup when the value of the input field is set to none', () => {
         comboBox.value = 'foo';
         comboBox.close();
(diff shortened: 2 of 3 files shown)


src/vaadin-combo-box.html
```diff
@@ -90,6 +91,8 @@
       _item-label-path="[[itemLabelPath]]"
       loading="[[loading]]">
     </vaadin-combo-box-dropdown-wrapper>
+
+    <iron-media-query query="(min-width: 600px)" query-matches="{{_desktopMode}}"></iron-media-query>
```
jouni: Device width is not a really detecting the input method, is it? :)

I don’t like the fact that we need a new dependency for this (iron-media-query). We should detect touch devices, whether the button was clicked with a finger or a touch device.

web-padawan:
Reviewed 5 of 5 files at r1.

---

*[src/vaadin-combo-box.html, line 95 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/620#-L6LUX5y1DHiNB1qzJ0x-r1-95:-L6LvR-GbTsDrswq3kIs:b-e5rcmv) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/47cf2df9ee96b2cb40356a5221b102b62dd25d0a/src/vaadin-combo-box.html#L95)):*
<details><summary><i>Previously, jouni (Jouni Koivuviita) wrote…</i></summary><blockquote>

Device width is not a really detecting the input method, is it? :)

I don’t like the fact that we need a new dependency for this (iron-media-query). We should detect touch devices, whether the button was clicked with a finger or a touch device.
</blockquote></details>

BTW, we use `420px` in dropdown-menu and date-piker, and also `750px` for "wide" in context-menu.  If we are not inclined to re-use our device detector as a separate element or mixin, let's at least align the numbers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/620)*

web-padawan:
:lgtm:

---

Reviewed 2 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/620#-:-L6MBD7kIsQnDWJ7N__t:bnfp4nl)*

tomivirkki:
:lgtm:

---

Reviewed 2 of 5 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/620#-:-L6MR9Xyc7Nzol2EAAu5:bnfp4nl)*

jouni:
The vaadin-device-detector in context-menu is a rather odd one. I feel like we should get rid of that, or at least extract it out as a separate utility mixin.

web-padawan:
@jouni +1 to the mixin, at least to detect touch device / IE / Safari  etc. Something like `DeviceAwareMixin` re-used wherever we need it.

tomivirkki:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-combo-box/620)*

Issue #680, opened 2018-07-22
The vaadin-combo-box does not render when the value for item-label-path is null


gatanaso:
### Description
The `vaadin-combo-box` does not render when the value for the specified `item-label-path` is `null`.

### Expected outcome
The `vaadin-combo-box` should gracefully handle those cases and render empty label(s) for those items.

### Actual outcome
<img width="485" alt="combo-box-error" src="https://user-images.githubusercontent.com/15094658/43045870-fdb15dda-8dc8-11e8-8fac-092a13896999.png">

### Steps to reproduce

1. Add a `vaadin-combo-box` element in the page and set the `item-label-path` property.
2. Set the items and ensure that at least one of the item objects has a value of `null` for the specified property that is used as the `item-label-path`.
3. Open the page in a web browser.
4. The `vaadin-combo-box` does not render on the page

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari

gatanaso:
PR merged, closing issue.

PR #681, opened 2018-07-22
add null handling for item label properties

gatanaso:
This PR fixes the issue (#680) of the `vaadin-combo-box` failing to render when an item property used as a label path has a `null` value.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/681)

### DIFF
--- a/src/vaadin-combo-box-dropdown-wrapper.html
+++ b/src/vaadin-combo-box-dropdown-wrapper.html
@@ -259,6 +259,8 @@ This program is available under Apache License Version 2.0, available at https:/
         let label = item ? this.get(this._itemLabelPath, item) : undefined;
         if (label === undefined) {
           label = item ? item.toString() : '';
+        } else if (label === null) {
+          label = '';
         }
         return label;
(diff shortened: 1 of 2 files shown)


gatanaso:
___
*[src/vaadin-combo-box-dropdown-wrapper.html, line 262 at r1](https://reviewable.io/reviews/vaadin/vaadin-combo-box/681#-LI4oSYkDGtCRHoLyRcF:-LI4wM3i62xeJRkkbFm5:bp9riqh) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/061a1f37b42f15889ccb79e3353f0c2a51de85f7/src/vaadin-combo-box-dropdown-wrapper.html#L262)):*

I feel that the correct code here should be like in the line above, set item.toString if item exists like it does with undefined.
</blockquote></details>

Thanks for the review. I have updated the PR.

YuriyVaadin:
Rebase on current master, should help with passing the build.

PR #705, opened 2018-08-24
Cleanup unused code and unnecessary DOM measurements

web-padawan:
Note: there are two changes which could be _potentially_ breaking:

- removal of `touch-device` attribute from `vaadin-combo-box-item` - __but__ we don't use it anymore, instead CSS media queries ` @media (pointer: coarse) ` are used both in Lumo and Material
- removal of `IronA11yKeysBehavior` from `vaadin-combo-box-light` - __but__ it is not used by component itself, neither by `vaadin-time-picker`, and anyone who wants to extending it can still use `Polymer.mixinBehaviors` himself.

Furthermore, both are not documented parts of the public API, e. g. `touch-device` attribute is not documented anywhere. So, we can consider removing those safely in minor release.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/705)

### DIFF
--- a/src/vaadin-combo-box-dropdown-wrapper.html
+++ b/src/vaadin-combo-box-dropdown-wrapper.html
@@ -5,9 +5,7 @@ This program is available under Apache License Version 2.0, available at https:/
 -->
 
 <link rel="import" href="../../polymer/polymer-element.html">
-<link rel="import" href="../../polymer/lib/utils/debounce.html">
 <link rel="import" href="../../iron-list/iron-list.html">
-<link rel="import" href="../../iron-resizable-behavior/iron-resizable-behavior.html">
 <link rel="import" href="vaadin-combo-box-item.html">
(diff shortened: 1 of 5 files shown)


tomivirkki:
___
*[src/vaadin-combo-box-mixin.html, line 17 at r1](https://reviewable.io/reviews/vaadin/vaadin-combo-box/705#-LKefDIl7lXdNKfcSo1C:-LKefDIl7lXdNKfcSo1D:bf0vo6a) ([raw file](https://github.com/vaadin/vaadin-combo-box/blob/c3161073e46ccd5af27c953323c71eaaad1853c0/src/vaadin-combo-box-mixin.html#L17)):*
> ```HTML
>   window.Vaadin = window.Vaadin || {};
> 
>   const TOUCH_DEVICE = (() => {
> ```

This produces a globally available "TOUCH_DEVICE"

PR #730, opened 2018-10-07
Fix readonly state's a11y

samiheikki:
Readonly input should not have role combobox or aria-expanded (true or false) attribute

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/730)

### DIFF
--- a/src/vaadin-combo-box.html
+++ b/src/vaadin-combo-box.html
@@ -292,7 +292,8 @@ This program is available under Apache License Version 2.0, available at https:/
              */
             disabled: {
               type: Boolean,
-              value: false
+              value: false,
+              observer: '_disabledChanged'
             },
 
             /**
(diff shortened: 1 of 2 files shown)


samiheikki:
screenshot 2018-11-25 at 12 14 52
This implementation is based on the output of Google Lighthouse A11Y test as shown in the image.

The "Learn more" link points to here: https://dequeuniversity.com/rules/axe/2.2/aria-allowed-attr?application=lighthouse but that does not really have any more details.

tomivirkki:
After reading https://github.com/validator/validator/issues/324, I think the correct approach would be to move the `role="combobox"` to a parent node. Some of the requirements described [here](https://www.w3.org/TR/wai-aria-1.1/#combobox) are also not fulfilled; a PR that fixes the component to properly align with the spec would be optimal.

samiheikki:
@tomivirkki Updated the PR to use vaadin-combo-box with `role` combobox and `aria-expanded` attribute

samiheikki:
Rebased.

rolfsmeds:
Closing as remaining a11y issues will be fixed in P3 versions.

PR #744, opened 2018-10-31
Reuse the same placeholder instance

tomivirkki:
The current master creates new placeholder instances for every unloaded item in the `filteredItems` array when a `dataProvider` is used. Optimize memory use by re-using the same placeholder instance for each pending item.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/744)


tomivirkki:
Added the suggested test. Ping @yuriyvaadin

PR #750, opened 2018-11-30
Postpone overlay stamping to first opening

### LINKED ISSUE #748: Scripting for ComboBox takes too much time.
### Description
We have problems with UI changes from server taking too much time to update on client side.
We are using a solution where we from the server at times adds big groups of vaadin components.
 
For example in one of our action the server calls for the creation of ~70 components. 

I did some test to see what was eating time on the client side.

In one test I added 72 ComboBox (12 visible and 60 hidden).

This took:
Scripting: 4525ms
Rendering: 629ms
Painting: 37ms

This is not okay, we need these times to come down.

### Example 
https://gist.github.com/brp-helena/2885920dc037b87c20273b630529d60b
That gist link contains a patch for bookstore-starter-flow, this is the same test code i have used to get the numbers above.

### Steps to reproduce
1. Add this patch to bookstore-starter or vaadin-starter
2. After login go to http://localhost:8080/object
3. Use chromes "Performance" under DevTools to the record the time from when "Load Object" is pressed.

### Browsers Affected
- [x ] Chrome

pekam:
This is a performance improvement.

When attaching a lot of combo boxes at once, the initial render can take a lot of time. Most of the time is spent on executing logic related to the overlay. This functionality is not needed until the overlay is opened, and it is wasted effort for any combo box on the page that the user doesn't open.

As a trade-off the first opening takes a bit longer. But the bigger problem was the huge delay caused by initializing a lot of combo boxes at once.

I ran a couple of tests by attaching 100 combo boxes, without and with this improvement. The rendering time improved enormously:
- Chrome: 1,960ms -> 249ms (down by 87%)
- Edge: 79,250ms -> 2,820ms (down by 96%)

Fixes #748

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/750)

### DIFF
--- a/src/vaadin-combo-box-dropdown-wrapper.html
+++ b/src/vaadin-combo-box-dropdown-wrapper.html
@@ -27,7 +27,7 @@ This program is available under Apache License Version 2.0, available at https:/
       }
     </style>
 
-    <template is="dom-if" if="[[_firstOpened]]">
+    <template id="dropdown-template" is="dom-if" if="[[_renderDropdown]]">
       <vaadin-combo-box-dropdown id="dropdown"
           hidden="[[_hidden(_items.*, loading)]]"
           position-target="[[positionTarget]]"
@@ -173,12 +173,13 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       _openedChanged(opened, items, loading) {
-        if (!this._firstOpened) {
+        if (!this.$.dropdown) {
           if (opened) {
-            this._firstOpened = true;
             this._initDropdown();
           }
-          else return;
+          else {
+            return;
+          }
         }
         // Do not attach if no items
         // Do not dettach if opened but user types an invalid search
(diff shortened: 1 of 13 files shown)


moderndeveloperllc:
@pekam Have you been able to quantify the difference for a single dropdown? < 100ms might not be a problem, but if there is a noticeable delay, that could be an issue for the 99% of us who don't have dozens of combo boxes per page.

pekam:
@moderndeveloperllc That's a valid concern. I tested the time it takes to open the overlay for the first time:
- Chrome: 44ms -> 63ms
- Edge: 2,230ms -> 2,400ms

Although this seems quite acceptable, the delay can be more noticeable on a less powerful machine (and for example on IE11). However, all of this added delay is reduced from the initial rendering time.

moderndeveloperllc:
@pekam Good to know. 63ms shouldn't be too bad for most users. Wow, is Edge really that slow, or it that from a VM (Browserstack, etc)? 

FYI, a somewhat decent paper on delay perception can be [found here](http://www.tactuallabs.com/papers/howMuchFasterIsFastEnoughCHI15.pdf). It's reasonably recent (2015) and has a good set of references to other papers. IMHO user tolerance of delay is getting smaller as hardware gets faster and faster. The old "100ms is fast enough" adage is from the early 90's when user expectations were rather mild and no longer applies in many cases.

src/vaadin-combo-box-dropdown-wrapper.html
```diff
@@ -170,33 +173,57 @@
...
-        if (/Trident/.test(navigator.userAgent)) {
-          this._scroller.setAttribute('unselectable', 'on');
-        }
+      _initDropdown() {
+        const template = this.shadowRoot.querySelector('#dropdown-template');
+        template.if = true;
+        template.render();
+
+        this.$.dropdown = this.shadowRoot.querySelector('#dropdown');
+        this.$.dropdown.$.overlay = this.$.dropdown.shadowRoot.querySelector('#overlay');
```
yuriy-fix: Nit: there should not be a need in this one, as it is defined in the scope of the `dropdown` reference set before.

src/vaadin-combo-box-dropdown-wrapper.html
```diff
@@ -12,48 +12,51 @@
...
+    <!-- Stamping the content is postponed until the combo box is opened for the first time -->
+    <template id="dropdown-template" is="dom-if">
+      <vaadin-combo-box-dropdown id="dropdown"
+          hidden="[[_hidden(_items.*, loading)]]"
+          position-target="[[positionTarget]]"
+          on-template-changed="_templateChanged"
+          on-position-changed="_setOverlayHeight"
+          theme="[[theme]]">
+        <template>
+          <style>
```
web-padawan: This doesn't seem to work in ShadyDOM, style tag is there in Edge and does not seem to be processed.

web-padawan: OK, looks like styles are indeed applied, although style tag is there in markup. Interesting.
For now we can leave it as it is, as long as the tests are passing.

src/vaadin-combo-box-dropdown-wrapper.html
pekam: So it seems. Removed.

PR #763, opened 2019-01-31
Fix issues when itemLabelPath and itemValuePath are undefined

### LINKED ISSUE #717: Setting itemLabelPath to undefined should produce an error message
### Description
If you (accidentally) set `itemLabelPath` to undefined, the popup can never be opened. Even if you change the value later on, the popup will never open. No error message or exception will be printed either.

### Expected outcome
The popup is opened with either the default `toString` value or empty values. An error message is printed 

### Actual outcome
Overlay does not open

### Live Demo
https://glitch.com/edit/#!/thunder-barracuda?path=index.html:21:0

niiyeboah:
Fixes #717

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/763)

### DIFF
--- a/src/vaadin-combo-box-dropdown-wrapper.html
+++ b/src/vaadin-combo-box-dropdown-wrapper.html
@@ -42,7 +42,7 @@ This program is available under Apache License Version 2.0, available at https:/
                   on-click="_onItemClick"
                   index="[[__requestItemByIndex(item, index)]]"
                   item="[[item]]"
-                  label="[[getItemLabel(item)]]"
+                  label="[[getItemLabel(item, _itemLabelPath)]]"
                   selected="[[_isItemSelected(item, _selectedItem, _itemIdPath)]]"
                   renderer="[[renderer]]"
                   role$="[[_getAriaRole(index)]]"
@@ -327,8 +327,9 @@ This program is available under Apache License Version 2.0, available at https:/
        * Gets the label string for the item based on the `_itemLabelPath`.
        * @return {String}
        */
-      getItemLabel(item) {
-        let label = item ? this.get(this._itemLabelPath, item) : undefined;
+      getItemLabel(item, itemLabelPath) {
+        itemLabelPath = itemLabelPath || this._itemLabelPath;
+        let label = item && itemLabelPath ? this.get(itemLabelPath, item) : undefined;
         if (label === undefined || label === null) {
           label = item ? item.toString() : '';
         }
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -159,7 +159,8 @@ This program is available under Apache License Version 2.0, available at https:/
          */
         itemLabelPath: {
           type: String,
-          value: 'label'
+          value: 'label',
+          observer: '_itemLabelPathChanged'
         },
 
         /**
@@ -220,7 +221,7 @@ This program is available under Apache License Version 2.0, available at https:/
         '_filteredItemsChanged(filteredItems.*, itemValuePath, itemLabelPath)',
         '_templateOrRendererChanged(_itemTemplate, renderer)',
         '_loadingChanged(loading)',
-        '_selectedItemChanged(selectedItem)',
+        '_selectedItemChanged(selectedItem, itemLabelPath)',
         '_toggleElementChanged(_toggleElement)'
       ];
     }
@@ -398,7 +399,7 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _getItemValue(item) {
-      let value = item ? this.get(this.itemValuePath, item) : undefined;
+      let value = item && this.itemValuePath ? this.get(this.itemValuePath, item) : undefined;
       if (value === undefined) {
         value = item ? item.toString() : '';
       }
@@ -623,8 +624,14 @@ This program is available under Apache License Version 2.0, available at https:/
       }
     }
 
+    _itemLabelPathChanged(itemLabelPath, oldItemLabelPath) {
+      if (typeof itemLabelPath !== 'string') {
+        console.error('You should set itemLabelPath to a valid string');
+      }
+    }
+
     _filterChanged(filter, itemValuePath, itemLabelPath) {
-      if (filter === undefined || itemValuePath === undefined || itemLabelPath === undefined) {
+      if (filter === undefined) {
         return;
       }
       if (this.items) {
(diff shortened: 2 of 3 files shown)


test/using-object-values.html
```diff
@@ -259,6 +264,45 @@
...
+
+        const selectItemComboboxUndefinedPaths = (index) => {
+          // simulates clicking on the overlay items, but it more reliable in tests.
+          comboboxUndefinedPaths.$.overlay.dispatchEvent(new CustomEvent('selection-changed', {
+            detail: {item: comboboxUndefinedPaths.items[index]}
+          }));
+        };
+
+        beforeEach(() => {
+          error = sinon.stub(console, 'error');
```
tomivirkki: If you're stubbing `console.error` on `beforeEach`, I'd suggest restoring it in `afterEach`

tomivirkki: Or optionally do the stubbing in the test cases directly

src/vaadin-combo-box-dropdown-wrapper.html
```diff
@@ -328,7 +328,7 @@
        * @return {String}
        */
       getItemLabel(item) {
-        let label = item ? this.get(this._itemLabelPath, item) : undefined;
+        let label = item && this._itemLabelPath ? this.get(this._itemLabelPath, item) : undefined;
```
tomivirkki: This is better than before but the intended logic still doesn't work correctly.
1. Open "Filtering with a Remote Data Source" (It uses actual objects as items)
1.1. Open the combo box once so the items are listed
2. Set `itemLabelPath` to undefined (Now this produces an error message but the labels are not updated (to the intended toString() output))
3. Set `itemLabelPath` to "link", which is a valid item subproperty -> The error goes away but the items still show old values

src/vaadin-combo-box-mixin.html
```diff
@@ -623,10 +624,29 @@
       }
     }
 
+    _itemLabelPathChanged(itemLabelPath, oldItemLabelPath) {
```
tomivirkki: There should already be functionality somewhere that refreshes the item labels since they're updated for example when the items change. No need to write new logic for it but instead make sure that the existing logic gets invoked also on `itemLabelPath` change

niiyeboah: Done

src/vaadin-combo-box-mixin.html
```diff
@@ -623,10 +624,21 @@
...
+        console.error('You should set itemLabelPath to a valid string');
+      }
+
+      if (this.selectedItem) {
+        this._inputElementValue = this._getItemLabel(this.selectedItem);
+      }
+    }
+
     _filterChanged(filter, itemValuePath, itemLabelPath) {
-      if (filter === undefined || itemValuePath === undefined || itemLabelPath === undefined) {
```
web-padawan: Why was this line changed? Especially, `itemValuelPath` change seems unrelated.
Same as above: please revert the change to this line in case if new logic does not require it.

src/vaadin-combo-box-mixin.html
```diff
@@ -742,7 +754,7 @@
     }
 
     _itemsOrPathsChanged(e, itemValuePath, itemLabelPath) {
-      if (e.value === undefined || itemValuePath === undefined || itemLabelPath === undefined) {
```
web-padawan: Same as above: please revert the change to this line in case if new logic does not require it.

src/vaadin-combo-box-mixin.html
```diff
@@ -759,7 +771,7 @@
     }
 
     _filteredItemsChanged(e, itemValuePath, itemLabelPath) {
-      if (e.value === undefined || itemValuePath === undefined || itemLabelPath === undefined) {
```
web-padawan: Same as above: please revert the change to this line in case if new logic does not require it.

src/vaadin-combo-box-mixin.html
```diff
@@ -623,10 +624,21 @@
...
+        this._inputElementValue = this._getItemLabel(this.selectedItem);
+      }
+    }
+
     _filterChanged(filter, itemValuePath, itemLabelPath) {
-      if (filter === undefined || itemValuePath === undefined || itemLabelPath === undefined) {
+      if (filter === undefined) {
         return;
       }
+
```
web-padawan: nit: please remove newline, so the diff would be cleaner.

src/vaadin-combo-box-mixin.html
web-padawan: Discussed in person: new test is failing without this change. OK to include it.

src/vaadin-combo-box-mixin.html
web-padawan: Discussed in person: new test is failing without this change. OK to include it.

src/vaadin-combo-box-mixin.html
web-padawan: Discussed in person, OK to include this change.

src/vaadin-combo-box-mixin.html
```diff
@@ -623,10 +624,21 @@
       }
     }
 
+    _itemLabelPathChanged(itemLabelPath, oldItemLabelPath) {
+      if (typeof itemLabelPath !== 'string') {
+        console.error('You should set itemLabelPath to a valid string');
+      }
+
+      if (this.selectedItem) {
+        this._inputElementValue = this._getItemLabel(this.selectedItem);
```
tomivirkki: This logic also already exists [here](https://github.com/vaadin/vaadin-combo-box/blob/master/src/vaadin-combo-box-mixin.html#L692) and it has some additional conditions to it also. Would be cleaner to reuse that logic instead of rewriting something similar here. So the fix would be to change the observer params to `'_selectedItemChanged(selectedItem, itemLabelPath)'`

I also noticed that the following code in the test suite produces different results. Can you check it's not a regression from this PR:

```
combobox.value = null;
combobox.itemLabelPath = 'custom';

console.log(combobox.inputElement.value); // "empty-string-custom"
```

```
combobox.itemLabelPath = 'custom';
combobox.value = null;

console.log(combobox.inputElement.value); // ""
```

niiyeboah: I have confirmed that this behaviour was there before my changes. If you change the `itemLabelPath` after the `value` change then because the `value` is set to empty string [here](https://github.com/vaadin/vaadin-combo-box/blob/adfe7b783dab90079f92a9ec283cc2427badf498/src/vaadin-combo-box-mixin.html#L679), the input element changes to the empty string value in the items, but if you change the `value` to null after the `itemLabelPath` change then the input element value is set to the empty string value [here](https://github.com/vaadin/vaadin-combo-box/blob/adfe7b783dab90079f92a9ec283cc2427badf498/src/vaadin-combo-box-mixin.html#L683).

niiyeboah: The `_selectedItemChanged(selectedItem, itemLabelPath)` change is done.

Issue #767, opened 2019-02-08
Investigate Edge performance issues [2d]


web-padawan:
From the internal feedback: 

> ComboBox (Flow) with lazy loading is _incredibly_ slow in Edge 17 and 18
However, even normal combo boxes and the web component seems slower in Edge 17 than in IE11. In particular, typing into the box is very laggy.

We recently delayed rendering of the overlay until first opening. Now we should make it open faster.

One example problem: the `position-changed` event is dispatched too often (3 times when opening): https://github.com/vaadin/vaadin-combo-box/blob/79826d7c46d99c7f9d1a343f162e9ffd9c824067/src/vaadin-combo-box-dropdown.html#L250 
Also, overlay repositioning logic can be refactored and moved to vaadin-combo-box-dropdown-wrapper

web-padawan:
Started by investigating the sequence of `_setPosition` calls. This method is invoked multiple times but even with memoization of the results of `getBoundingClientRect` performance is poor.

### Chrome

1. set position when opened set to true: 
2. set position from iron-resize: 95
3. set position from notify resize: 96
4. set position after animation frame: 98

### Edge

1. set position when opened set to true: 
2. set position from iron-resize: 1731
3. set position from notify resize: 1731
4. set position after animation frame: 1763

So, it looks like the most of work happens between `opened` set to true and `iron-resize` event. This logic involves `_onOpened` method, which calls `Polymer.flush` to render first 3 items based on hardcoded metrics, and then calls `updateViewportBoundaries` to render whole list.

So, the assumption from the initial issue appears to be not correct. I continue investigating to understand what can we do to make the mentioned part faster in Edge.

Haprog:
The related improvements (#776) are now included in [v4.2.6](https://github.com/vaadin/vaadin-combo-box/releases/tag/v4.2.6)

PR #776, opened 2019-02-20
Replace dom-if with disable upgrade, delay scroll position adjustment

web-padawan:
This makes combo-box opening faster at least in my virtual machine, although it still takes 1 second to adjust scroll after overlay opening.

The `disable-upgrade` change is also intended to make the initial opening faster, and especially avoid one extra `Polymer.flush` call caused by `render` but it does not seem to help much by itself.

Connected to #767

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/776)

### DIFF
--- a/src/vaadin-combo-box-dropdown-wrapper.html
+++ b/src/vaadin-combo-box-dropdown-wrapper.html
@@ -12,51 +12,50 @@ This program is available under Apache License Version 2.0, available at https:/
 
 <dom-module id="vaadin-combo-box-dropdown-wrapper">
   <template>
-    <!-- Stamping the content is postponed until the combo box is opened for the first time -->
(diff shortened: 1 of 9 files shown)


moderndeveloperllc:
@web-padawan Wasn't the lazy attach a fix for when there are lots of combo boxes on a page? I seem to remember a previous issue where having all the boxes stamp on initialization caused pretty bad slowdowns on IE(?)

web-padawan:
The `disable-upgrade` also prevents stamping internals of the overlay, so there's only small difference. So, it should not affect initial page rendering.

Haprog:
This fix is now included in [v4.2.6](https://github.com/vaadin/vaadin-combo-box/releases/tag/v4.2.6)

PR #782, opened 2019-03-12
test: fix flaky tests due to scroll timing change

web-padawan:
Attempting to fix the flakyness introduced after 41f8709 due to the delay for initial scroll adjustment.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/782)


Haprog:
Looks like only one visual test failed above ("default-tests-material default [firefox]") but I think that's unrelated to this issue.

web-padawan:
@Haprog thanks, restarted and it failed again. I will investigate and fix that to get build green.

PR #785, opened 2019-03-13
Close overlay before setting selectedItem

### LINKED ISSUE #783: DataProvider is called after filtering and selecting an item
Steps to reproduce:
1. Set a filter
2. Select an item

The filter is changed to empty string, but the `opened` property is still true (even though the overlay is closing).
Because the filter has changed and the combo box thinks it's still opened, `clearCache()` gets called and the first page gets requested: https://github.com/vaadin/vaadin-combo-box/blob/41f870912bdd872e3376b090e863641928568a24/src/vaadin-combo-box-data-provider-mixin.html#L101

pekam:
Otherwise, the observer of selectedItem resets the filter, which causes
the dataProvider to be called, because combo box still thinks it's open.

Fix #783

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/785)

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -866,12 +866,12 @@ This program is available under Apache License Version 2.0, available at https:/
         return;
       }
 
-      if (this.selectedItem !== e.detail.item) {
-        this.selectedItem = e.detail.item;
-      }
-
       if (this.opened) {
         this.close();
+      }
+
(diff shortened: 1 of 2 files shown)


Haprog:
This fix is now included in [v4.2.6](https://github.com/vaadin/vaadin-combo-box/releases/tag/v4.2.6)

PR #794, opened 2019-03-20
feat: add clear-button-visible API from <vaadin-text-field>

### LINKED ISSUE #786: Integrate text-field's clearButtonVisible API
Remove the existing dedicated clear button and use the included text-field's clearButtonVisible API instead.
Fix the tests to use the new functionality.
Add demo, update visual tests...

The new functionality will make the clear button hidden by default = breaking change

Prototype https://github.com/vaadin/vaadin-combo-box/tree/clear-button-visible

Haprog:
BREAKING CHANGE: The clear button is now hidden by default. To make it visible, set the `clear-button-visible` attribute or use the `clearButtonVisible` property.

Fixes #786

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/794)


Haprog:
Looks like 3 keyboard navigation related tests are failing on Safari only. Can reproduce locally.

One lazy loading test is failing on old iPhone simulator. Seems like not related to my changes? Needs to be checked.

I also forgot to update Firefox screenshots for visual tests. Will push those soon.

Haprog:
Now the only test failures are 4 lazy-loading tests in "iOS Simulator iphone 9.3" (for Polymer 3 version only)... 🤔

```
iOS Simulator iphone 9.3 ✖ test/lazy-loading.html » lazy loading
    » combo-box » dataProvider » when open » should clear filter on value clear
    » combo-box-light » dataProvider » when open » should clear filter on value clear
    » combo-box-light » dataProvider » when open » should not request on value change
    » combo-box-light » itemIdPath » should support deep paths
```

Haprog:
I installed iOS 9.3 simulator and ran all tests on simulated iPhone 4s with iOS 9.3: only one test fails "filtering.html > filtering items > setting items when opened > should properly display all items in the selector" but that is because that test fails always when viewport height is less than 500px (also on desktop browsers) and mobile Safari viewport height on iPhone 4s is apparently something like 480px. Also ran them with iPhone 5 (iOS 9.3) and the result was the same, only this test failing due to not so tall viewport.

Then I also ran the tests with iPhone 6 Plus (iOS 9.3) simulator and all tests pass. So it seems the failure only happens a bit randomly in SauceLabs.

src/vaadin-combo-box-mixin.html
```diff
@@ -352,8 +352,9 @@
     _onClick(e) {
       this._closeOnBlurIsPrevented = true;
 
+      const clearButton = this.$.input ? this.$.input.$.clearButton : this._clearElement;
```
platosha: BTW the `_clearElement`-style accessors were introduced as an abstraction from the DOM implementation. The actual DOM location of the clear button depends on the component implementation and is different in regular and `-light` components. With this change, we handle that difference inside the shared mixin code, which is slightly more dirty.

Since the `vaadin-combo-box-light` does not need to access `this.$.input`, it would rather make sense to keep using `_clearElement` throughout the mixin code, and apply the change in the regular `vaadin-combo-box.html` instead.

Of couse, would be best to remove all `clearButton` references from the mixin altogether. But if that’s not possible, would perhaps make sense to keep using `_clearElement` as an abstracted reference.

src/vaadin-combo-box-mixin.html
```diff
@@ -352,8 +352,9 @@
...
       const path = e.composedPath();
-      if (path.indexOf(this._clearElement) !== -1) {
+      if (path.indexOf(clearButton) !== -1) {
         this._clear();
```
platosha: BTW Clearing is already implemented in the text-field itself. That suggests that, ideally, we should remove the clear implementation from here altogether, I think.

I do understand that it might be not possible to do right now, so it’s fine to clean that up later. Let me kindly ask for a TODO comment here to remind us about that though.

test/vaadin-combo-box-light.html
```diff
@@ -174,6 +174,50 @@
 
         expect(spy.called).to.be.true;
       });
+
+      describe('custom clear-button', () => {
```
platosha: BTW Not sure if that feature is useful now:

- With native inputs, there are native clear buttons
- If `<vaadin-text-field>` is used inside `<vaadin-combo-box-light>`, the clear button from the text field could be reused

platosha: Now that we have a major version bump after this, would it make sense to drop this feature too, perhaps?

src/vaadin-combo-box-mixin.html
Haprog: Yeah I agree some parts of the code are now a bit more dirty. But so seems to be much of the combo box code. 😄

I was thinking about setting `_clearElement` to `this.$.input` in `vaadin-combo-box.html` but that would have required some other changes. Probably something like introducing a new property e.g. `_customClearElement` which would be used to handle the logic needed for a custom clear button and using `_clearElement` only for the cases which should be handled the same way for both the custom clear button and the clear button coming from `vaadin-text-field`.

src/vaadin-combo-box-mixin.html
Haprog: Yeah I think it would be good if it worked in a way that text field just clears its value which gets passed on, but I think it didn't work so simply right now so that this was needed to make it work for now. I can add a comment about it tomorrow.

test/vaadin-combo-box-light.html
Haprog: > With native inputs, there are native clear buttons

Really? 😲

> If <vaadin-text-field> is used inside <vaadin-combo-box-light>, the clear button from the text field could be reused

Good point, but I think if you need `vaadin-combo-box-light.html` it's also likely that you want to use some other input than `<vaadin-text-field>`. This could be supported too though.

I don't really like our current implementation of supporting a custom clear button much, but I think it's still a good feature. I wouldn't probably like to drop the feature completely, but it might make sense to try to rethink how this use case is supported (if we can simplify our code by doing so). Maybe by providing a bit different API to be able to do it with `<vaadin-combo-box-light>` even if it would require a bit more code from the developer using `<vaadin-combo-box-light>` as long as it's not too difficult to use. But I think that would be a separate enhancement.

We could discuss this tomorrow.

Haprog:
Reran the Travis job that failed, and this time it succeeded.

src/vaadin-combo-box-mixin.html
Haprog: I was able to refactor this so that the mixin only references `_clearElement` instead of 
 `this.$.input`.

src/vaadin-combo-box-mixin.html
Haprog: I tried testing if it would be possible to get this working without having at least some clear logic in either the mixin or both vaadin-combo-box and vaadin-combo-box-light, but it seems that's not really possible. We need at least the focus call after clearing, even if it would be possible to get the value to update automatically.

I tried separating the custom clear button behaviour to vaadin-combo-box-light.html and having the vaadin-text-field clear button related logic in vaadin-combo-box.html, but I wasn't able to make it any cleaner than it was so I left this bit as is in the mixin code. I didn't add a TODO comment since I'm not sure what could be done about it.

src/vaadin-combo-box-light.html
```diff
@@ -74,16 +80,18 @@
...
      * @mixes Vaadin.ComboBoxMixin
      * @mixes Vaadin.ThemableMixin
      * @mixes Vaadin.ThemePropertyMixin
      */
     class ComboBoxLightElement extends
-      Vaadin.ThemePropertyMixin(
-        Vaadin.ThemableMixin(
-          Vaadin.ComboBoxDataProviderMixin(
-            Vaadin.ComboBoxMixin(Polymer.Element)))) {
+      Vaadin.ControlStateMixin(
```
platosha: Not a good idea to have `Vaadin.ControlStateMixin` here, where the host element does not need to delegate focus into shadow DOM, the input is in the light DOM and can receive focus naturally.

This brings a lot of workarounds over things that should work natively already, and this is risky.

In particular, because the mixin makes the focus element focusable, when tabbing into `<vaadin-combo-box-light><input></vaadin-combo-box-light>`, there is an extra step with a focus frame around `<vaadin-combo-box-light>`.

test/vaadin-combo-box-light.html
platosha: > Really? 😲

I mean `<input type="search">` 🤓 

Fine by me to keep the feature.

src/vaadin-combo-box-light.html
```diff
@@ -150,6 +163,20 @@
           this.inputElement[this._propertyForValue] = value;
         }
       }
+
+      /**
+       * Focusable element used by vaadin-control-state-mixin
+       */
+      /*
```
platosha: Kind reminder to remove the commented code.

test/vaadin-combo-box-light.html
```diff
@@ -174,6 +174,118 @@
...
+          // Get the viewport relative position of the element
+          const rect = elem.getBoundingClientRect();
+          if (rect.width === 0 || rect.height === 0) {
+            return;
+          }
+          const x = Math.ceil(rect.left);
+          const y = Math.ceil(rect.top);
+          // Get the element which would be targeted, when the user
+          // tries to click on this position
+          let target = elementFromPointDeep(x, y, elem.ownerDocument);
```
platosha: Super great that you’ve found a way to test this. 👍

src/vaadin-combo-box-light.html
Haprog: Oops. Thanks for noticing. Will do. 😄

Haprog: Done.

PR #796, opened 2019-03-21
fix: change event not called when item selected via click

### LINKED ISSUE #792: on-change callback not working on version 4.2.6.
### Description
I have a combo box with an `on-change` event handler. The event handler only gets called when I clear the selection, not changing the selection. It happens in version 4.2.6 when using in Polymer 2.
```html
<vaadin-combo-box
  theme="small"
  hidden="[[_hideEnvironmentSwitch]]"
  label="Change Environment"
  on-change="_changeEnvironment"
  items="[[_environments]]"
  value="{{_selectedEnvironment}}">
</vaadin-combo-box>
```

```js
_changeEnvironment() {
  debugger
  if (this._selectedEnvironment) {
    // some logic
  }
  this._toggleEnvironmentSelector();
}
```

Haprog:
Fixes #792

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/796)

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -872,6 +872,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
       if (this.selectedItem !== e.detail.item) {
         this.selectedItem = e.detail.item;
+        this._detectAndDispatchChange();
       }
     }
 
--- a/test/selecting-items.html
+++ b/test/selecting-items.html
@@ -247,6 +247,14 @@
         combobox.value = 'foo';
         combobox.close();
       });
+
+      it('should fire when selecting an item via click', () => {
+        combobox.open();
+        const firstItem = combobox.$.overlay._selector.querySelector('vaadin-combo-box-item');
+        fire('click', firstItem);
+
+        expect(changeSpy.callCount).to.equal(1);
+      });
     });
   });
 


src/vaadin-combo-box-mixin.html
```diff
@@ -872,6 +872,7 @@
 
       if (this.selectedItem !== e.detail.item) {
         this.selectedItem = e.detail.item;
+        this._detectAndDispatchChange();
```
web-padawan: adding this line here means that the following one has stopped working:
https://github.com/vaadin/vaadin-combo-box/blob/f3d2c44f65580e34a2b4c4087c95eb555f47484f/src/vaadin-combo-box-mixin.html#L599

Can we now remove that one and see if all the tests still pass?

Haprog: No, that one should still work for other cases, but here the problem (regression cause) was that now `close()` is called before the `selectedItem` is set so there are no changes at that point. See:
```js
      if (this.opened) {
        this.close();
      }
      if (this.selectedItem !== e.detail.item) {
        this.selectedItem = e.detail.item;
        this._detectAndDispatchChange();
      }
```

Haprog: But I will check if that other one could maybe be not needed anymore. I'll try running the tests without it, but I think it's still needed for when the overlay is closing is triggered in other ways.

Haprog: If that other `this._detectAndDispatchChange();` is removed then 3 tests fail.
image

web-padawan: Ok, that's what I suspected but needed to clarify that.

Haprog:
This fix will be part of the next patch release (presumably v4.2.7) which will probably be released today.

Haprog:
The fix is now in [v4.2.7](https://github.com/vaadin/vaadin-combo-box/releases/tag/v4.2.7).

PR #797, opened 2019-03-25
fix: bind value correctly when using attr-for-value

### LINKED ISSUE #778: Value gets reset when using attr-for-value different than "value"
### Description
Using a custom input element with a different attribute for value than `value` leads to the input being reset to `undefined` whenever the user starts typing.

Renaming the property to `value` makes it work (but that's of course difficult if one can't control the input element).

This seemed to be caused by the `ComboBoxMixin`, which hardcodes the attribute name to `value`: https://github.com/vaadin/vaadin-combo-box/blob/master/src/vaadin-combo-box-mixin.html#L607
The line sets the value of the combo box always to `undefined` (because the `value` property doesn't exist).

### Expected outcome
Value from the input box is not overwritten and used by the combo-box.

### Actual outcome
Value is reset to `undefined` whenever the user starts typing.

### Live Demo
https://glitch.com/edit/#!/peaceful-consonant?path=index.html:25:0

### Steps to reproduce
Create a custom input element that exposes its value as a property that is called different than `value`:
```javascript
import {PolymerElement, html} from '@polymer/polymer/polymer-element';
import '@polymer/iron-input';

class MyInput extends PolymerElement {
	static get template() {
		return html`
		<iron-input bind-value="{{search}}">
			<input>
		</iron-input>`;
	}

	static get properties() {
		return {
			search: {
				notify: true,
				type: String,
			},
		};
	}
}
window.customElements.define('my-input', MyInput);
```

Use this custom input element with the Vaadin combo-box:
```html
<dom-bind id="vaadin">
	<template>
		<vaadin-combo-box-light attr-for-value="search" items="[[items]]">
			<my-input class="input"></my-input>
		</vaadin-combo-box-light>
	</template>
</dom-bind>

<script>
	var scope = document.querySelector('#vaadin');
	scope.items = ['apples', 'pears', 'bananas'];
</script>
```

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

Haprog:
When using `attr-for-value` different than "value", the data binding of value wasn't fully working. This only happened to work with the `iron-input` example we have ("Replacing the Input" example) because `iron-input` also has a "value" property. When using a custom input with `attr-for-value` other than "value" the user was unable to change the value of the input by interacting with it via the keyboard. This is now also working.

Fixes #778

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/797)

### DIFF
--- a/test/my-input.html
+++ b/test/my-input.html
@@ -1,0 +1,36 @@
+<!--
+  This custom element is needed for testing the "attr-for-value" feature of
+  <vaadin-combo-box-light>
+-->
+<link rel="import" href="../../polymer/polymer-element.html">
+<script>
+  class MyInput extends Polymer.Element {
+    static get is() {
+      return 'my-input';
+    }
+
+    static get template() {
+      const {html} = Polymer;
+      return html`
+      <style>
+        :host {
+          display: inline-block;
+        }
+      </style>
+      <iron-input id="input" bind-value="{{customValue}}">
+        <input>
+      </iron-input>
+      `;
+    }
+  
+    static get properties() {
+      return {
+        customValue: {
+          type: String,
+          notify: true
+        }
+      };
+    }
+  }
+  window.customElements.define('my-input', MyInput);
+</script>
--- a/test/my-input.html
+++ b/test/my-input.html
@@ -3,34 +3,32 @@
   <vaadin-combo-box-light>
 -->
 <link rel="import" href="../../polymer/polymer-element.html">
-<script>
-  class MyInput extends Polymer.Element {
-    static get is() {
-      return 'my-input';
-    }
+<dom-module id="my-input">
+  <template>
+    <style>
+      :host {
+        display: inline-block;
+      }
+    </style>
+    <iron-input id="input" bind-value="{{customValue}}">
+      <input>
+    </iron-input>
+  </template>
+  <script>
+    class MyInput extends Polymer.Element {
+      static get is() {
+        return 'my-input';
+      }
 
-    static get template() {
-      const {html} = Polymer;
-      return html`
-      <style>
-        :host {
-          display: inline-block;
-        }
-      </style>
-      <iron-input id="input" bind-value="{{customValue}}">
-        <input>
-      </iron-input>
-      `;
+      static get properties() {
+        return {
+          customValue: {
+            type: String,
+            notify: true
+          }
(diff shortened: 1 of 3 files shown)


Haprog:
Looks like the new test doesn't work on IE11. And probably something weird happening with the Polymer 3 conversion of the new `MyInput` test element since it complains about `html is not a function` which probably relates to this part:
```js
const {html} = Polymer;
return html`...`;
```

test/my-input.html
```diff
@@ -0,0 +1,34 @@
...
+-->
+<link rel="import" href="../../polymer/polymer-element.html">
+<dom-module id="my-input">
+  <template>
+    <style>
+      :host {
+        display: inline-block;
+      }
+    </style>
+    <iron-input id="input" bind-value="{{customValue}}">
```
platosha: Missing import for `<iron-input>` component dependency here.

test/vaadin-combo-box-light.html
```diff
@@ -142,6 +151,73 @@
       });
     });
 
+    describe('vaadin-combobox-light-attr-for-value', () => {
+      let comboBox, customInput, ironInput, nativeInput;
+
+      function whenIronInputReady(ironInput, callback) {
```
web-padawan: can we use `ironInput._observer.flush()` instead of this callback?
see the observer to init slotted input detection [here](https://github.com/PolymerElements/iron-input/blob/5f9099a045554e9fe666bca614b22c18f77247e5/iron-input.html#L174)

Haprog: I'll check it out, thanks for the suggestion.

test/my-input.html
Haprog: Added, thanks.

test/vaadin-combo-box-light.html
Haprog: Seems to work, and simplifies the code quite a bit. Thanks!

Haprog: Update pushed.

Haprog:
This fix is now in [v4.2.8](https://github.com/vaadin/vaadin-combo-box/releases/tag/v4.2.8).

PR #800, opened 2019-04-05
perf: defer overlay positioning to improve typing in Edge

platosha:
Fixes vaadin/vaadin-combo-box-flow#245

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/800)

### DIFF
--- a/src/vaadin-combo-box-dropdown.html
+++ b/src/vaadin-combo-box-dropdown.html
@@ -225,6 +225,9 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       _setPosition(e) {
+        if (this.hidden) {
+          return;
+        }
         if (e && e.target) {
           const target = e.target === document ? document.body : e.target;
           const parent = this.$.overlay.parentElement;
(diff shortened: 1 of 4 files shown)


pekam:
This has a bug: 
1. Filter the overlay to show zero items and wait for 500ms
2. Change the filter to show some items again
3. Notice how the overlay is translated to a wrong position for 500ms, and then the position is fixed

The problem is actually not related to the added debouncing, but the `_setPosition` function, where `overlay.getBoundingClientRect()` return 0 for `left` and `top` when the overlay is hidden, which messes up the calculations.

I pushed a possible fix, which seems to work fine, to another branch:
https://github.com/vaadin/vaadin-combo-box/compare/perf/edge-filtering...perf/edge-filtering-2

Haprog:
Some lazy loading tests are failing on iOS (few seem to be randomly failing on CI but not reproducible locally). One test failure is reproducible locally, but not 100% of the time so it seems a bit flaky. This fix still seems to work for the issue so we're merging this now and we'll try to fix the tests later (probably in next sprint).

Issue #801, opened 2019-04-29
Test failing in iOS 9.3 in Travis and simulator


web-padawan:
The following test started to be flaky after 395e706

```
iOS Simulator iphone 9.3 ✖ test/lazy-loading.html » lazy loading » combo-box-light » dataProvider » should be invoked with size set
  expected dataProvider to have been called exactly once, but it was called twice
    dataProvider({ filter: "", page: 0, pageSize: 50 }, function callback() {}) at http://localhost:8081/components/@vaadin/[secure]/src/[secure]-data-provider-mixin.js:83:34
    dataProvider({ filter: "", page: 1, pageSize: 50 }, function callback() {}) at http://localhost:8081/components/@vaadin/[secure]/src/[secure]-data-provider-mixin.js:83:34
        dataProvider({ filter: "", page: 0, pageSize: 50 }, function callback() {}) at http://localhost:8081/components/ at vaadin/[secure]/src/[secure]-data-provider-mixin.js:83:34
        dataProvider({ filter: "", page: 1, pageSize: 50 }, function callback() {}) at http://localhost:8081/components/ at vaadin/[secure]/src/[secure]-data-provider-mixin.js:83:34
                                   <unknown> at   AssertionError at /components/chai/chai.js:5270:0
                                   <unknown> at           assert at /components/chai/chai.js:265:0
                                   <unknown> at        <unknown> at /components/sinon-chai/lib/sinon-chai.js:91:0
                                   <unknown> at      addProperty at /components/chai/chai.js:3961:0
                                   <unknown> at        <unknown> at lazy-loading.html:146:0
```

Haprog:
These are also related and flaky in Travis

```
iOS Simulator iphone 11.3 ✖ test/lazy-loading.html » lazy loading » combo-box
  » dataProvider » when open » should not request on value change
  AssertionError: expected dataProvider to not have been called

iOS Simulator iphone 11.3 ✖ test/lazy-loading.html » lazy loading » combo-box-light
  » dataProvider » when open » should not request on value change
  AssertionError: expected dataProvider to not have been called

iOS Simulator iphone 9.3 ✖ test/lazy-loading.html » lazy loading » combo-box-light
  » dataProvider » when open » should not request on value change
```

Haprog:
Still seems to be a problem in CI at least with these tests:
```
[...]

PR #803, opened 2019-05-03
fix: clear button related issue with vaadin-combo-box-light

### LINKED ISSUE #802: Value isn't cleared properly when using clear-button-visible API of vaadin-text-field within vaadin-combo-box-light
### Description
By default `vaadin-combo-box` uses `vaadin-text-field` and supports using the `clear-button-visible` API. `vaadin-combo-box-light` works with any input element and support using a custom clear button element (`class="clear-button"`) but if you use a `vaadin-text-field` with `clear-button-visible` set to true within `vaadin-combo-box-light`, the click of the clear button doesn't properly propagate the value clearing to the combo box.

So in this case if you click the clear button of `vaadin-text-field` it correctly clears the value from the native `<input>` and from `vaadin-text-field` itself, but `vaadin-combo-box-light` doesn't pick up this change and the value of `vaadin-combo-box-light` doesn't match the value of `vaadin-text-field` anymore.

For example `vaadin-time-picker` basically uses a `vaadin-combo-box-light` with `vaadin-text-field` (though it is a custom class that extends `vaadin-text-field`) enabling the use of `clear-button-visible` on the `vaadin-text-field`.

We should make it easy to use the clear button available in `vaadin-text-field` also when it's used with `vaadin-combo-box-light`.

I think this works in `vaadin-combo-box` only because it uses proper data binding to bind the combo box value with `vaadin-text-field` value, but `vaadin-combo-box-light` can't properly data bind to the input that is given to it via a slot.

This can be worked around by the developer who uses `vaadin-text-field` with `vaadin-combo-box-light` by data binding the values together in a parent component template (or with `<dom-bind>`).

It would be good if we can make this work automatically without requiring the above "workaround". But if there is no easy way to do this or if we don't want to do this, then at least we should document it so that if you use `vaadin-text-field` within `vaadin-combo-box-light` you should also make sure to data bind their values together (at least if you use the `clear-button-visible` API).

### Expected outcome
`comboBoxLight.value === comboBoxLight.inputElement.value`
`comboBoxLight.value` === `''`
`comboBoxLight.inputElement.value` === `''`

### Actual outcome
`comboBoxLight.value !== comboBoxLight.inputElement.value`
`comboBoxLight.value` === `'Helium'` (or something else, whatever was set as the value before clicking the clear button)
`comboBoxLight.inputElement.value` === `''`

### Live Demo
Can be reproduced in the "Configuring Icons" demo at https://cdn.vaadin.com/vaadin-combo-box/5.0.0/demo/#combo-box-styling-demos if you remove the element with `class="clear-button"` and add `clear-button-visible` to the `vaadin-text-field`.

### Steps to reproduce
1. Select/inspect the `vaadin-combo-box-light` element (of the "Configuring Icons" demo) in dev tools.
2. `$0.querySelector('.clear-button').remove();`
3. `$0.inputElement.clearButtonVisible = true;`
4. `$0.value = 'Helium';`
5. Click the clear button (or run `$0.inputElement.$.clearButton.click()`)
6. Don't yet unfocus the element (leave dropdown open)
7. Notice that `$0.value` === `'Helium'` but `$0.inputElement.value` === `''` and you can see that Helium item is still shown as selected in the dropdown. You can see the property values easiest if you create a live expression in Chrome dev tools so you will see the changes in those properties live (if you click on dev tools it will already unfocus the element, close the dropdown and values will be updated again).

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

Tested on Chrome but probably affects all browsers.

Haprog:
Fixes #802

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/803)

### DIFF
--- a/src/vaadin-combo-box-light.html
+++ b/src/vaadin-combo-box-light.html
@@ -115,6 +115,7 @@ This program is available under Apache License Version 2.0, available at https:/
       constructor() {
         super();
         this._boundInputValueChanged = this._inputValueChanged.bind(this);
+        this.__boundInputValueCommitted = this.__inputValueCommitted.bind(this);
       }
 
       ready() {
@@ -141,13 +142,23 @@ This program is available under Apache License Version 2.0, available at https:/
         this._setInputElement(this.querySelector(cssSelector));
         this._revertInputValue();
         this.inputElement.addEventListener('input', this._boundInputValueChanged);
+        this.inputElement.addEventListener('change', this.__boundInputValueCommitted);
         this._preventInputBlur();
       }
 
       disconnectedCallback() {
         super.disconnectedCallback();
         this.inputElement.removeEventListener('input', this._boundInputValueChanged);
+        this.inputElement.removeEventListener('change', this.__boundInputValueCommitted);
         this._restoreInputBlur();
+      }
+
+      __inputValueCommitted(e) {
+        // Detect if the input was cleared (e.g. by clicking the clear button on a vaadin-text-field)
+        // and propagate the value change to combo box value immediately.
+        if (this._inputElementValue === '') {
+          this._clear();
+        }
       }
 
       get _propertyForValue() {
--- a/test/vaadin-combo-box-light.html
+++ b/test/vaadin-combo-box-light.html
@@ -74,6 +74,14 @@
     </template>
   </test-fixture>
 
+  <test-fixture id="combobox-light-vaadin-text-field">
+    <template>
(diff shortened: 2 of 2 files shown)


Haprog:
Only few flaky tests (#801) are failing on iOS (which are also failing on `master`). So this MR should be ok to merge after passing review.

PR #808, opened 2019-05-14
Stop enter keyboard event, add test

yuriy-fix:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-combo-box/808)

### DIFF
--- a/test/keyboard.html
+++ b/test/keyboard.html
@@ -193,6 +193,13 @@
         expect(comboBox.opened).to.equal(false);
       });
 
+      it('should stop propagation of the keyboard enter event', () => {
+        const keydownSpy = sinon.spy();
+        document.addEventListener('keydown', keydownSpy);
+        enter();
+        expect(keydownSpy.called).to.be.false;
+      });
+
       it('should not close the overlay with enter when custom values are not allowed', () => {
         filter('foobar');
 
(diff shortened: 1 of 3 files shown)


test/keyboard.html
```diff
@@ -193,6 +193,13 @@
         expect(comboBox.opened).to.equal(false);
       });
 
+      it('should stop propagation of the keyboard enter event', () => {
+        const keydownSpy = sinon.spy();
+        document.addEventListener('keydown', keydownSpy);
+        enter();
+        expect(keydownSpy.called).to.be.false;
```
Haprog: I would prefer to use the Sinon spy assertion helpers. E.g.
```js
expect(keydownSpy).to.not.have.been.called;
```
(I think this should work though I haven't used this exact format)

You can see some examples in: https://github.com/vaadin/vaadin-combo-box/blob/master/test/lazy-loading.html#L77

And docs at: https://www.chaijs.com/plugins/sinon-chai/

Haprog: Though this is not a blocker, and otherwise LGTM.

yuriy-fix: Updated.

Haprog:
@yuriy-fix btw now the linked issue can only be seen via ZenHub. Please remember to mention the related issue number directly in PR or commit too, so also externals can see what is the related issue and then it can also be picked up by `magi` to release notes.

I'll edit the PR description to link to the issue now.

PR #841, opened 2019-09-17
Fix custom value bug

### LINKED ISSUE #839: Saving a repetitive data after pressing ESC and then TAB
### Description
When you write a data that is already in the ComboBox item, after pressing ESC and then tab, the data is saved in the ComboBox. 

### Expected outcome
We expect that the data should be reverted to the default value of the ComboBox and not be saved in the ComboBox at all.

### Actual outcome
A repetitive data is saved in the ComboBox.

### Steps to reproduce
1. Set the allow-custom-value for combo-box. 
2. Type something that already exists in the ComboBox item and then presses ESC and then presses tab. 
3. Instead of reverting to the default value or null value, the data is saved again in the ComboBox.

shadikhani:
Fix #839

### DIFF
--- a/test/vaadin-combo-box.html
+++ b/test/vaadin-combo-box.html
@@ -231,6 +231,17 @@
             comboBox.$.overlay._selector.querySelector('vaadin-combo-box-item').click();
             expect(spy).to.not.have.been.called;
           });
+
+          it('should not be fired when existing item is entered and overlay is closed', () => {
+            const spy = sinon.spy();
+            comboBox.addEventListener('custom-value-set', spy);
+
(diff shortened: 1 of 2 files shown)


test/vaadin-combo-box.html
```diff
@@ -384,8 +395,8 @@
           comboBox.items = ['bar', 'baz'];
           comboBox.open();
           expect(
-            comboBox.$.overlay._selector.querySelector('vaadin-combo-box-item')
```
tomivirkki: You'll need to revert these whitespace changes to make the lint check pass

PR #844, opened 2019-09-25
fix: don't set custom value when event not fired

pekam:
This part should have been included inside the if-block added in #841.

It doesn't even make sense to check if `e.defaultPrevented`, when the event is not fired, and there's nothing to prevent it.


pekam:
> This re-introduces [#841 (review)](https://github.com/vaadin/vaadin-combo-box/pull/841#pullrequestreview-289855816)

Fixed in a way that should cover both cases.

PR #846, opened 2019-10-01
updated pageSizeObserver to check for negative values

### LINKED ISSUE #845: Setting a negative value for 'pageSize' does not throw an error
### Description
Currently it is possible to set a negative value for `pageSize`. This causes errors when using lazy loading as it incorrectly tries to use the negative value.

This is an issue with the `pageSizeObserver` which only does the following check:
`Math.floor(pageSize) !== pageSize || pageSize === 0`
which does not take into consideration negative values, albeit the error message indicating as such: `pageSize value must be an integer > 0`
(for more details see: https://github.com/vaadin/vaadin-combo-box/blob/fab032fd601fb9a93db1959a61fb312628872ac6/src/vaadin-combo-box-data-provider-mixin.html#L202)

### Expected outcome
The combo box throws an error, same as it does when setting non number values or 0.

### Actual outcome
The negative value is set and no error is thrown to indicate that an incorrect value was set.

gatanaso:
Added a check for negative values to the `pageSizeObserver`, such that it will not allow negative values to be set for `pageSize`. Similarly to how setting non-number values or 0 throws an exception, setting a negative value should throw one as well.

fixes #845

### DIFF
--- a/src/vaadin-combo-box-data-provider-mixin.html
+++ b/src/vaadin-combo-box-data-provider-mixin.html
@@ -200,7 +200,7 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _pageSizeChanged(pageSize, oldPageSize) {
-      if (Math.floor(pageSize) !== pageSize || pageSize === 0) {
+      if (Math.floor(pageSize) !== pageSize || pageSize === 0 || pageSize < 0) {
         this.pageSize = oldPageSize;
         throw new Error('`pageSize` value must be an integer > 0');
       }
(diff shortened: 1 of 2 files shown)


src/vaadin-combo-box-data-provider-mixin.html
```diff
@@ -200,7 +200,7 @@
     }
 
     _pageSizeChanged(pageSize, oldPageSize) {
-      if (Math.floor(pageSize) !== pageSize || pageSize === 0) {
+      if (Math.floor(pageSize) !== pageSize || pageSize === 0 || pageSize < 0) {
```
tomivirkki: Isn't `pageSize === 0 || pageSize < 0` the same as `pageSize < 1` which makes sense (pagesize shouldn't be less than 1) and reads better

gatanaso: Valid point 😄 I will update accordingly, thanks for the feedback @tomivirkki 🙏

PR #853, opened 2020-01-16
Add autoOpenDisabled property

### LINKED ISSUE #863: Implement autoOpenDisabled
When `comboBox.autoOpenDisabled = true`:
1. Focusing and typing in field doesn’t open the overlay
2. Clicking on icon opens the overlay
3. Up / Down keys opens the overlay

niiyeboah:
When `comboBox.autoOpenDisabled = true`:
- Focusing and typing in field doesn’t open the overlay
- Clicking on icon opens the overlay
- Up / Down keys opens the overlay

Closes #863

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -36,6 +36,11 @@ This program is available under Apache License Version 2.0, available at https:/
         },
 
         /**
+         * Set true to prevent the overlay from opening automatically.
+         */
+        autoOpenDisabled: Boolean,
+
+        /**
          * Set to true to disable this element.
          */
         disabled: {
@@ -366,7 +371,7 @@ This program is available under Apache License Version 2.0, available at https:/
       } else if (path.indexOf(this.inputElement) !== -1) {
         if (path.indexOf(this._toggleElement) > -1 && this.opened) {
           this.close();
-        } else {
+        } else if (path.indexOf(this._toggleElement) > -1 || !this.autoOpenDisabled) {
           this.open();
         }
       }
@@ -638,7 +643,7 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _filterFromInput(e) {
-      if (!this.opened && !e.__fromClearButton) {
+      if (!this.opened && !e.__fromClearButton && !this.autoOpenDisabled) {
         this.open();
       }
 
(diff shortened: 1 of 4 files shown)


tomivirkki:
The visual tests aren't passing. Might be a font rendering change issue. If that's the case, update reference screenshots with `npm i -q --no-save gemini@^4.0.0 gemini-sauce gemini-polyserve && node_modules/.bin/gemini update test/visual`

niiyeboah:
I updated the visual tests. Not sure what the issue was, maybe font rendering like you suggested.

hongduc-phan:
Hi  Niiyeboah. If you set comboBox.autoOpenDisabled = true, and you type Option1 or Option2, It has 2 issues:
1. We type Option one ( for example), the value prop is not updated.
Screenshot 2020-01-31 at 15 17 02

2. there is not any suggestion for user.( not sure it is bug)

src/vaadin-combo-box-mixin.html
```diff
@@ -352,7 +357,7 @@
       } else if (path.indexOf(this.inputElement) !== -1) {
         if (path.indexOf(this._toggleElement) > -1 && this.opened) {
           this.close();
-        } else {
+        } else if (path.indexOf(this._toggleElement) > -1 || !this.autoOpenDisabled) {
           this.open();
```
hongduc-phan: not sure but IMOP, it should be refactored like 
`if ((path.indexOf(this._toggleElement) > -1) {
if (this.opened) {
this.close();
}
else if (!this.autoOpenDisabled) {
this.open();
}
`

niiyeboah: Hey I tried to edit your comment so the code was clearer, hope you don't mind

niiyeboah: In the original code, the outer `if` is checking for the presence of the `inputElement` and the inner ones are checking for the `toggleButton`, I don't think your refactor accounts for all that logic. There might be a way to make this cleaner but my approach was to add on in a way that changed as little as possible.

hongduc-phan: Thank you for editing my comment. :) yes. :) I understand your mean.

tomivirkki:
All the PR's related to this feature should have a related GH issue. Serhii created one for [`vaadin-date-picker`](https://github.com/vaadin/vaadin-date-picker/issues/704) but the rest of the repos also need one. This requirement (all PR's should have a related GH issue) is also stated in the internal contribution guidelines document.

web-padawan:
@niiyeboah added one change to refactor code so now we separate `close` (normal combo-box) and `commit` (combo-box being edited with `autoOpenDisableld`). The tests still pass.

niiyeboah:
> @niiyeboah added one change to refactor code so now we separate `close` (normal combo-box) and `commit` (combo-box being edited with `autoOpenDisableld`). The tests still pass.

@web-padawan Thanks! Just checked the code, separating the functions like that makes a lot of sense 👍

PR #855, opened 2020-01-23
Add RTL specific styles

### LINKED ISSUE #854: Add RTL specific styles
Fix the margins for item / item in the overlay.
Create tests for fading effect that should be on the left side in RTL mode.

yuriy-fix:
Fixes #854


theme/lumo/vaadin-combo-box-dropdown-styles.html
```diff
@@ -84,6 +84,14 @@
           transform: rotate(360deg);
         }
       }
+
+      /* RTL specific styles */
+
+      :host([loading]) [part~="loader"] {
+        left: auto;
+        margin-left: 0;
+        margin-right: auto;
```
web-padawan: This probably needs to also override `margin-inline-start` and `margin-inline-end`.

yuriy-fix: Good point. Weird that we have used those as they are not supported by ie11. Updated.

yuriy-fix:
The problem was coming from `DirMixin`: https://github.com/vaadin/vaadin-element-mixin/pull/34

PR #856, opened 2020-01-25
Fixes maxOverlayHieght(height of scroller) #630

### LINKED ISSUE #630: [Question] _maxOverlayHeight() item limit + overlay height
When using 4.0.0-beta4, it appears that the maximum combo-box height is controlled in the `vaadin-combo-box-dropdown-wrapper.html` private method `_maxOverlayHeight`.

I have two questions regarding this:

1. `const minHeight = 116;` which the comment notes is the height of two items in the combo-box. Should this variable be hoisted to public so that it is configurable by end developers?

2. `const bottom` determines the min between the `window.innerHeight` and the math of `scrollHeight-scrollTop`. This has the side effect that if page has no scroll but has room for the combo box to be taller, it will default to 116px due to the math in the return statement `Math.max(bottom - targetRect.bottom - margin, minHeight) + 'px'`. A simple addition of a case to check for `bottom = 0` results in a better experience:

```
// Presuming the change `const bottom` to `let bottom`, when we have 
// no scroll on the page, we can expand beyond the min list count of two 
// to something more comfortable
if (bottom === 0) {
 bottom = window.innerHeight;
}
```

Note, the current documentation notes that `--vaadin-combo-box-overlay-max-height` will set the scroller overlay style, but the current code path in `_setOverlayHeight()` doesn't take that into account when setting `MaxHeight` for the dropdown overlay (and I'm not saying it should, it's just unclear as to what changes the current list height).

Current behavior:
image

After `bottom === 0` check:
image

Thoughts appreciated. Happy to submit a pull request if either 1 or 2 above make sense, or happy to close if I've simply missed a potential solution to the items viewable at default.

hongduc-phan:
Fixes #630


test/scrolling.html
```diff
@@ -45,6 +45,26 @@
         });
       });
 
+      describe('the height of scroller', () => {
+        beforeEach(() => {
```
web-padawan: This test also passes on master. In order to make a proper test you need to tweak `body` height.

Add this line to `beforeEach` to reproduce the case that causes the original issue:
```js
document.body.style.height = '100px';
```

Then also remember to unset it back so that other tests won't be affected:
``` js
afterEach(() => {
  document.body.style.height = null;
});
```

hongduc-phan: done

PR #858, opened 2020-01-27
test: split visual tests for item template

### LINKED ISSUE #857: Update dropdown item template visual tests
### Description

The visual tests in `dropdown.html` incorrectly contain two combo-boxes.
This has been unnoticed at least since 379ef3b where Material was added.

Here is what the current screenshot looks like:

visual test

We should discuss whether this is a bug or not, and if so, create another issue for it.

This issue is about updating visual test so maybe we have `dropdown-template.html`
It would be more reliable than what we currently have with two combo-boxes.

web-padawan:
Fixes #857


web-padawan:
Closing in favor of #855 where the visual tests are updated.

PR #876, opened 2020-04-29
feat: custom overlay width

### LINKED ISSUE #529: Allow customizing overlay width
Expose a property like `--vaadin-combo-box-width` that can be used to configure the overlay width. 

Would be helpful especially in situations where the dropdown contents are wider than the actual input field. 
screen_shot_2017-09-18_at_5 12 11_pm

Haprog:
This makes it possible to easily set a custom overlay width (either bigger or smaller than the input) with a CSS custom property `--vaadin-combo-box-overlay-width` on the `<vaadin-combo-box>` element. The component will automatically propagate this custom property to the overlay element so it is easy to target a specific combo box without needing to use a custom `theme` name for the combo box.

This is a simplistic implementation that does not adjust the overlay positioning so the issue mentioned here is still relevant: https://github.com/vaadin/vaadin-combo-box/issues/529#issuecomment-605956167
> the popup is always left-aligned with the text field. This means if your combo is near the right-hand edge of the window, the popup is cropped.
That should be opened as a separate issue once this is merged.

Fixes #529.

### DIFF
--- a/demo/combo-box-basic-demos.html
+++ b/demo/combo-box-basic-demos.html
@@ -154,6 +154,33 @@
       </template>
     </vaadin-demo-snippet>
 
+    <h3>Custom overlay width</h3>
+    <vaadin-demo-snippet id="overlay-width" when-defined="vaadin-combo-box">
+      <template preserve-content>
+        <vaadin-combo-box label="Narrow overlay"></vaadin-combo-box>
+        <vaadin-combo-box label="Wide overlay"></vaadin-combo-box>
+        <script>
+          window.addDemoReadyListener('#overlay-width', function(document) {
+            const combos = document.querySelectorAll('vaadin-combo-box');
+            // If you don't need IE 11 support you can set the overlay width also with an inline
+            // style attribute like: style="--vaadin-combo-box-overlay-width: 130px".
+            combos[0].style.setProperty('--vaadin-combo-box-overlay-width', '130px');
+            combos[1].style.setProperty('--vaadin-combo-box-overlay-width', '330px');
+            if (window.ShadyCSS) {
+              window.ShadyCSS.styleSubtree(combos[0], {'--vaadin-combo-box-overlay-width': '130px'});
+              window.ShadyCSS.styleSubtree(combos[1], {'--vaadin-combo-box-overlay-width': '330px'});
+            }
+
+            combos.forEach(function(comboBox) {
+              comboBox.items = Array.apply(0, Array(100)).map(function(_, i) {
+                return 'Item ' + (i + 1);
+              });
+            });
+          });
+        </script>
+      </template>
+    </vaadin-demo-snippet>
+
     <h3>Lazy loading</h3>
     <p>
       The <code>dataProvider</code> property can be assigned a function to provide
--- a/demo/combo-box-basic-demos.html
+++ b/demo/combo-box-basic-demos.html
@@ -154,33 +154,6 @@
       </template>
     </vaadin-demo-snippet>
 
-    <h3>Custom overlay width</h3>
-    <vaadin-demo-snippet id="overlay-width" when-defined="vaadin-combo-box">
-      <template preserve-content>
-        <vaadin-combo-box label="Narrow overlay"></vaadin-combo-box>
-        <vaadin-combo-box label="Wide overlay"></vaadin-combo-box>
-        <script>
-          window.addDemoReadyListener('#overlay-width', function(document) {
-            const combos = document.querySelectorAll('vaadin-combo-box');
-            // If you don't need IE 11 support you can set the overlay width also with an inline
-            // style attribute like: style="--vaadin-combo-box-overlay-width: 130px".
-            combos[0].style.setProperty('--vaadin-combo-box-overlay-width', '130px');
-            combos[1].style.setProperty('--vaadin-combo-box-overlay-width', '330px');
-            if (window.ShadyCSS) {
-              window.ShadyCSS.styleSubtree(combos[0], {'--vaadin-combo-box-overlay-width': '130px'});
-              window.ShadyCSS.styleSubtree(combos[1], {'--vaadin-combo-box-overlay-width': '330px'});
-            }
-
-            combos.forEach(function(comboBox) {
-              comboBox.items = Array.apply(0, Array(100)).map(function(_, i) {
-                return 'Item ' + (i + 1);
-              });
-            });
-          });
-        </script>
-      </template>
-    </vaadin-demo-snippet>
-
     <h3>Lazy loading</h3>
     <p>
       The <code>dataProvider</code> property can be assigned a function to provide
--- a/src/vaadin-combo-box-dropdown.html
+++ b/src/vaadin-combo-box-dropdown.html
@@ -249,8 +249,25 @@ This program is available under Apache License Version 2.0, available at https:/
         this._translateY = Math.round(this._translateY * _devicePixelRatio) / _devicePixelRatio;
         this.$.overlay.style.transform = `translate3d(${this._translateX}px, ${this._translateY}px, 0)`;
 
-        this.$.overlay.style.width = this.positionTarget.clientWidth + 'px';
         this.$.overlay.style.justifyContent = this.alignedAbove ? 'flex-end' : 'flex-start';
+
+        const customOverlayWidth = window.ShadyCSS ?
+          window.ShadyCSS.getComputedStyleValue(this, '--vaadin-combo-box-overlay-width') :
+          getComputedStyle(this).getPropertyValue('--vaadin-combo-box-overlay-width');
+
+        if (customOverlayWidth === '') {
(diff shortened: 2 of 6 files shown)


Haprog:
I'm not sure if the custom property name should be more generic so we could use the same property name in other similar components that have a dropdown overlay (like implementing the same feature for `<vaadin-select>`). Would it be better as `--vaadin-overlay-width` or `--vaadin-dropdown-width`?

There could also be a helper property so you could set it with an attribute as `overlay-width="123px"` which looks a bit nicer than setting a long custom property with inline `style` attribute. This could also translate nicely into Java API as `setOverlayWidth()`.

Haprog:
Some visual tests are failing the build. Taking a look.

src/vaadin-combo-box-dropdown.html
```diff
@@ -249,9 +249,26 @@
         this._translateY = Math.round(this._translateY * _devicePixelRatio) / _devicePixelRatio;
         this.$.overlay.style.transform = `translate3d(${this._translateX}px, ${this._translateY}px, 0)`;
 
-        this.$.overlay.style.width = this.positionTarget.clientWidth + 'px';
```
Haprog, line 252: This line is converted below to:
```js
const defaultOverlayWidth = this.positionTarget.clientWidth + 'px';
```

Haprog:
Some discussion about this PR https://vaadin.slack.com/archives/G6QQTHU8Z/p1588152064159400

Pekka suggested moving the demo under "Styling". Should I do it?

src/vaadin-combo-box-dropdown.html
```diff
@@ -249,9 +249,26 @@
...
         this.$.overlay.style.justifyContent = this.alignedAbove ? 'flex-end' : 'flex-start';
 
+        const customOverlayWidth = window.ShadyCSS ?
```
web-padawan, line 254: Please consider refactoring the code in two separate methods:

```js
_getCustomWidth() {
  return window.ShadyCSS ?
    window.ShadyCSS.getComputedStyleValue(this, '--vaadin-combo-box-overlay-width') :
    getComputedStyle(this).getPropertyValue('--vaadin-combo-box-overlay-width');
}

_setOverlayWidth() {
  const inputWidth = this.positionTarget.clientWidth + 'px';
  const customWidth = this._getCustomWidth();

  if (window.ShadyCSS) {
    window.ShadyCSS.styleSubtree(this.$.overlay, {
      '--vaadin-combo-box-overlay-width': customWidth,
      '--_vaadin-combo-box-overlay-default-width': inputWidth
    });
  } else {
    this.$.overlay.style.setProperty('--_vaadin-combo-box-overlay-default-width', inputWidth);

    if (customWidth === '') {
      this.$.overlay.style.removeProperty('--vaadin-combo-box-overlay-width');
    } else {
      this.$.overlay.style.setProperty('--vaadin-combo-box-overlay-width', customWidth);
    }
  }
}
```

theme/lumo/vaadin-combo-box-dropdown-styles.html
```diff
@@ -13,6 +13,7 @@
       }
 
       :host {
+        width: var(--vaadin-combo-box-overlay-width, var(--_vaadin-combo-box-overlay-default-width, auto));
```
web-padawan, line 16: Should this be in `src` rather than copying the same code to both themes?

demo/combo-box-basic-demos.html
```diff
@@ -154,6 +154,33 @@ <h3>Storing custom values</h3>
...
     </vaadin-demo-snippet>
 
+    <h3>Custom overlay width</h3>
+    <vaadin-demo-snippet id="overlay-width" when-defined="vaadin-combo-box">
+      <template preserve-content>
+        <vaadin-combo-box label="Narrow overlay"></vaadin-combo-box>
+        <vaadin-combo-box label="Wide overlay"></vaadin-combo-box>
+        <script>
+          window.addDemoReadyListener('#overlay-width', function(document) {
+            const combos = document.querySelectorAll('vaadin-combo-box');
```
web-padawan, line 164: Let's add `Array.from` to make sure `forEach` below works in al browsers.

Haprog: I was thinking about that but this demo page already has 2 older demos that also use `forEach()` on the result of `querySelectorAll()` directly and I tested that these demos work in IE11 so I thought it's fine.

But looking at it closer now, it looks like ShadyDOM patches `querySelectorAll()` to return an Array instead of NodeList, so that's why it works.

So I guess it's fine in our demos since they assume ShadyDOM anyway for IE11 and that's the only (still) relevant browser for us that doesn't natively [support Nodelist.forEach](https://caniuse.com/#feat=mdn-api_nodelist_foreach)? Also I don't see how `Array.from` would be better, since that is also not natively supported in IE11, but polyfilled by webcomponentsjs.

Haprog: Imo we only need `Array.from` in this kind of cases when we would use other array methods that are not available directly on [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) (or not polyfilled for IE11).

theme/lumo/vaadin-combo-box-dropdown-styles.html
Haprog: Yeah, good catch, since this is not Lumo or Material specific. I thought about it at some point, but wanted to avoid creating a new style module just for this.

I'm not sure if it can somehow cleanly be added directly to the custom element definition here https://github.com/vaadin/vaadin-combo-box/blob/3921cbf84f504128a7d3408288c753eef3ba5e91/src/vaadin-combo-box-dropdown.html#L50 (since it's just inheriting the vaadin-overlay template now).
I think it's probably better/cleaner to just make a small style module for this right above the class? I think I'll do that unless you have a better suggestion.

src/vaadin-combo-box-dropdown.html
Haprog: Thanks for the suggestion. This looks better. Though initially I intentionally left setting the native custom props out of an else block so that the real custom props would be set on even if you have ShadyCSS enabled in Chrome or other browsers. At least when just running the demos locally with `polymer serve` even Chrome has ShadyCSS enabled by default (I guess mostly for ApplyShim) so if your code depended on the real custom properties, it would not work.

But now that I looked a bit more into this I came up with a better check.

First I though I could [feature detect CSS custom property support](https://stackoverflow.com/questions/38012009/how-to-feature-detect-for-css-custom-properties) like `window.CSS && CSS.supports('color', 'var(--primary)')` but that also won't work right since it returns true in Chrome regardless of what polyfills are loaded and then I tested that the overlay size actually breaks with some polyfill combinations with only that check. With `wc-shimcssproperties` query param Chrome also needs to use `ShadyCSS.styleSubtree`.

Then I looked at what `ShadyCSS` contains and it seems `ShadyCSS.nativeCss` is a boolean that tells whether CSS custom properties are native or shimmed so that seems like the correct check to work in all scenarios. I'll make an update using this.

Haprog:
```js
if (window.ShadyCSS && !window.ShadyCSS.nativeCss) {
  // use ShadyCSS.styleSubtree
} else {
  // set native custom props
}
```
This seems to work nicely in both IE11 and Chrome (with different polyfill combinations).

On Chrome I tested with:
[demo/](http://0.0.0.0:3000/components/vaadin-combo-box/demo/#combo-box-basic-demos)
[demo/?wc-shadydom](http://0.0.0.0:3000/components/vaadin-combo-box/demo/?wc-shadydom#combo-box-basic-demos) (ShadyDOM without shimming custom props)
[demo/?wc-shadydom&wc-shimcssproperties](http://0.0.0.0:3000/components/vaadin-combo-box/demo/?wc-shadydom&wc-shimcssproperties#combo-box-basic-demos) (ShadyDOM and shimming custom props)

Note `wc-shimcssproperties` without `wc-shadydom` doesn't seem to work at all (all styles are broken) so I guess `wc-shimcssproperties` is only meant to be used with `wc-shadydom`.

I also briefly tried `wc-ce` with some of the others but that didn't seem to make a difference.

Haprog: This also makes it so that `ShadyCSS.styleSubtree` is not unnecessarily called on browsers that might have ApplyShim loaded but don't shim custom props (like Chrome by default when running this demo locally.

Haprog: Updated the code. PTAL.

theme/lumo/vaadin-combo-box-dropdown-styles.html
Haprog: Updated the code. PTAL.

demo/combo-box-basic-demos.html
web-padawan: Thanks for clarification. No need to change then, let's keep it as it is.

PR #879, opened 2020-05-14
Preserve and propagate dir to overlay and items

yuriy-fix:
Connected to https://github.com/vaadin/vaadin-time-picker/issues/145

### DIFF
--- a/test/vaadin-combo-box.html
+++ b/test/vaadin-combo-box.html
@@ -68,9 +68,11 @@
       describe('Direction', () => {
         it('should preserve and propagate dir to the dropdown overlay', () => {
           comboBox.setAttribute('dir', 'ltr');
+          document.documentElement.setAttribute('dir', 'rtl');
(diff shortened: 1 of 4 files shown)


yuriy-fix:
P.S. I can also create an `enhancement` ticket for preserving `dir` on nested elements and add a comment with the `// Workaround for issue`

test/vaadin-combo-box.html
```diff
@@ -65,6 +65,15 @@
         });
       });
 
+      describe('Direction', () => {
```
DiegoCardoso, line 68: If I got it correctly, this should override whenever was defined on the document level and set by `DirMixin`. Can we add a test for that? Like, if setting the document dir to RTL and `combo-box` to LTR, the overlay would have the one defined for the `combo-box`?

yuriy-fix: Have enhanced the test with setting `documentElement` `dir`

PR #885, opened 2020-06-03
test: update ref screenshot

DiegoCardoso:
[Visual tests](https://travis-ci.org/github/vaadin/vaadin-combo-box/jobs/694149882) is failing on `master` after #884 (even though it has just passed).


DiegoCardoso:
This is really weird. I have switched to Node 8.11 and re-ran the tests. It updated to the previous state and now I have a PR with an empty list of files. 😅
Test will likely pass, but `master` is still red.

PR #902, opened 2020-09-10
Keep the scroller position while scrolling lazily and with undefined size

mshabarov:
Fixes https://github.com/vaadin/vaadin-combo-box-flow/issues/386

### DIFF
--- a/src/vaadin-combo-box-data-provider-mixin.html
+++ b/src/vaadin-combo-box-data-provider-mixin.html
@@ -64,16 +64,6 @@ This program is available under Apache License Version 2.0, available at https:/
         /** @private */
         __placeHolder: {
           value: new Vaadin.ComboBoxPlaceholder()
-        },
-
-        /** @private
-         * Max item count in active range (items cache).
-         * This internal property reflects the 'MAX_RANGE_COUNT' in combo box
-         * connector.
-         */
-        __maxRangeCount: {
-          type: Number,
-          value: Math.max(this.pageSize * 2, 500)
         }
 
       };
@@ -138,13 +128,18 @@ This program is available under Apache License Version 2.0, available at https:/
         return true;
       }
 
+      // Max item count in active range (items cache).
+      // This local var reflects the 'MAX_RANGE_COUNT' in combo box
+      // connector.
+      const MAX_RANGE_COUNT = Math.max(this.pageSize * 2, 500);
+
       // We skip the page loading when the scrolling has been reset to 0
       // position after assigning a new items to iron-list, because otherwise
       // the connector's 'clearPageCallbacks' would be broken and the gaps in
       // items list occur. Checking items cache size here, since the gaps only
       // appear when the cache is fully populated and it should be possible for
       // sure to load the page at the very beginning
-      if (this.$.overlay.resetScrolling && this.filteredItems.length > this.__maxRangeCount) {
+      if (this.$.overlay.resetScrolling && this.filteredItems.length > MAX_RANGE_COUNT) {
         return false;
       }
 
--- a/src/vaadin-combo-box-data-provider-mixin.html
+++ b/src/vaadin-combo-box-data-provider-mixin.html
@@ -139,7 +139,7 @@ This program is available under Apache License Version 2.0, available at https:/
       // items list occur. Checking items cache size here, since the gaps only
       // appear when the cache is fully populated and it should be possible for
       // sure to load the page at the very beginning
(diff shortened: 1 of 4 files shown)


web-padawan (dismissed):

web-padawan:
I assume the fix can be tested in `ComboBoxFlow`, but it would be good to have a test for it here as well.

tomivirkki:
The scroll position should ideally be restored when only the `size` gets changed (not when the items is completely assigned a new value, for example on filter).

As it is, the suggested change causes the following UX regression (on filter the index is restored to a random position):

combo-box-index

Compare to the behavior in master:

combo-box-index-master

mshabarov:
> As it is, the suggested change causes the following UX regression (on filter the index is restored to a random position):

Good catch, thanks. Indeed, when we apply a filter, it may happen that the `_restoreScrollerPosition` method is invoked with the obsolete position and results in such a weird random scrolling. I fixed this by setting `oldScrollPosition` to 0 so as to avoid execution this scroll pos restore code.

> The scroll position should ideally be restored when only the size gets changed (not when the items is completely assigned a new value, for example on filter).

I noticed the methods `_getItems` and `_restoreScrollerPosition` are invoked only when the size is changed actually. If you use a defined size mode and just scrolling to the end of list, then it only invoked once at the beginning.

tomivirkki:
Just tried with the latest revision and it still works the same as before:

Kapture 2020-09-21 at 10 29 03

mshabarov:
I made a fix to address this regression with filtering and tested it with basic demo example using large size item array. Unit test has been modified as well, because previously it passed even with this regression. Please review.

src/vaadin-combo-box-data-provider-mixin.html
```diff
@@ -128,6 +128,21 @@
...
+      // connector.
+      const MAX_RANGE_COUNT = Math.max(this.pageSize * 2, 500);
+
+      // We skip the page loading when the scrolling has been reset to 0
+      // position after assigning a new items to iron-list, because otherwise
+      // the connector's 'clearPageCallbacks' would be broken and the gaps in
+      // items list occur. Checking items cache size here, since the gaps only
+      // appear when the cache is fully populated and it should be possible for
+      // sure to load the page at the very beginning
+      if (this.$.overlay.resetScrolling && this.size > MAX_RANGE_COUNT) {
```
tomivirkki, line 142: Is the MAX_RANGE_COUNT check needed for anything? I'd imagine we could always skip loading pages when the scrolling is being reset?

Also, this if-clause should have test cases (for both conditions).

Additional though: Would it be possible to have this logic inside the wrapper module instead so that as much as possible of the scroll reset logic would be in the same place? If the conditions apply, the wrapper could skip dispatching the "index-requested" event.

src/vaadin-combo-box-dropdown-wrapper.html
```diff
@@ -168,7 +199,41 @@
...
+          // new items size might be less than old scrolling position
+          this._selector.scrollToIndex(Math.min(items.length - 1, this._oldScrollerPosition));
+          this.resetScrolling = false;
+          // reset position to 0 again in order to properly handle the filter
+          // cases (scroll to 0 after typing the filter)
+          this._oldScrollerPosition = 0;
+        }
+      }
+
+      _notEmpty(items) {
```
tomivirkki, line 234: Given the signature of this function, you'd expect it to check if the provided array is not empty. But it also seems to rely on a state of a global `this._selector` variable, which is a bit unexpected.

Could change it to only check for the array: `return items && items.length`. If this new helper is added to the module, it could also be used [here](https://github.com/vaadin/vaadin-combo-box/blob/master/src/vaadin-combo-box-dropdown-wrapper.html#L184) and [here](https://github.com/vaadin/vaadin-combo-box/blob/master/src/vaadin-combo-box-dropdown-wrapper.html#L472)

tomivirkki (changes requested):

src/vaadin-combo-box-data-provider-mixin.html
mshabarov: You're right, I've removed MAX_RANGE_COUNT and it seems to work still. I assume my previous issue which I tried to resolve by this limitation has been resolved by another code change.

This condition has been moved to wrapper, as you advised. Better to have it there.

src/vaadin-combo-box-dropdown-wrapper.html
mshabarov: Done

tomivirkki (approved):

Issue #900, opened 2020-08-18
Valid item is not selected if ComboBox has a dataprovider and autoopendisabled


alvarezguille:
### Description
The input is cleared even though there's a match in the backend items.
Main problem seems to be that typing in the input does not trigger a data provider request unless the popup is opened.

### Live Demo
Can be reproduced in [v5.2.0 demo](https://cdn.vaadin.com/vaadin-combo-box/5.2.0/demo/#combo-box-basic-demos)

### Steps to reproduce
* Set autoOpenDisabled = true to the ComboBox in lazy loading demo
* Type a valid country name without opening the popup 
* Blur the field

### Expected outcome
Item is selected

### Actual outcome
Item is cleared

DiegoCardoso:
The solution proposed is to change the behavior of how data is fetched for combo-box regardless if `autoOpenDisabled` is on or off.

#### This is the user case that this issue should cover:

1. User start typing on combo-box field
1. Data is fetched based on the user input
1. _**User moves focus away from combo-box**_
1. Value inserted by the user **is not cleared** from the field while data is being fetched
1. When dataProvider returns with the response, then:
  	- if **there's a match**, set it as the combo-box value; or
  	- if **there's no match**, clear combo-box

User's feedback while data is being fetched should be handled by #901.

tomivirkki:
Related: https://github.com/vaadin/vaadin-combo-box/issues/919

Issue #775, opened 2019-02-15
Not selecting filtered option after tabbing out of element


fagnercarvalho:
### Description
The `vaadin-combo-box` element is not selecting the filtered option after tabbing out of the field.

### Expected outcome
The current value should be selected when tabbing out.

### Actual outcome
The current value is not selected.

### Live Demo
There is no live demo for this issue, sorry.
However I recorded a [short video showing the issue](https://1drv.ms/v/s!ApECU_YagH6sgZBOgyMgIDU9FPyVcw).

### Steps to reproduce
1. Put a `vaadin-combo-box` element in the page.
2. Open the page in a web browser.
3. Search for a valid option in the combo box  but don't select this option.
4. Tab out of the field.

### Browsers Affected

- [X] Chrome (only browser I tested)
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

rolfsmeds:
This would be a very useful feature for heavy data entry applications (not uncommon use for Vaadin), where speed is important. Having to press arrow-down and enter before tabbing to the next field might not sound like much, but the speed with which data entry users fill in forms can really make this a serious workflow bottleneck.

I would propose that this is provided as a setting on the ComboBox, since the described behavior might not always be desired.

web-padawan:
Works for me. Would be nice to see if at least a code snippet to reproduce the issue.

rolfsmeds:
Code is not relevant here -- we're referring to the default standard behavior of the CB.

User action currently required to select item:
1. Type into field to filter items
2. Arrow-down to focus desired item
3. Tab to blur field
-> Focused item is selected

Desired optional mode to select an item:
1. Type into field to filter **until only one item is available**
2. Tab to blur field
-> Only available item is selected

Alternately, this could even be:
1. Type into field to filter items
2. Tab to blur field
-> **First** available item is selected

web-padawan:
> Type into field to filter until only one item is available

I see. Now it only gets selected once you entered all the characters. Otherwise, it isn't.

rolfsmeds:
Proposal:
- Let's add an API to the combox called autoFocusFirstMatch
- When enabled, the topmost match is automatically focused when the filter text is changed and non-empty

mhosio:
Adding a simple workaround here that should enable the basic use case while waiting for the final implementation.

```
[...]

PR #961, opened 2020-10-21
Fix: set value on dataprovider callback

tulioag:
fixes: #900

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -552,7 +552,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
     /** @private */
     _closeOrCommit() {
-      if (this.autoOpenDisabled && !this.opened && !this.loading) {
+      if (!this.opened && !this.loading) {
         this._commitValue();
       } else {
         this.close();
@@ -666,7 +666,9 @@ This program is available under Apache License Version 2.0, available at https:/
         this.close();
       }
 
-      this._commitValue();
+      if (!this.loading) {
+        this._commitValue();
+      }
     }
 
     /** @private */
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -699,7 +699,7 @@ This program is available under Apache License Version 2.0, available at https:/
             this._selectItemForValue(customValue);
             this.value = customValue;
           }
-        } else if (!this.opened && itemsMatchedByLabel.length == 1) {
+        } else if (!this.allowCustomValue && !this.opened && itemsMatchedByLabel.length == 1) {
           this.value = this._getItemValue(itemsMatchedByLabel[0]);
         } else {
           this._inputElementValue = this.selectedItem ? this._getItemLabel(this.selectedItem) : (this.value || '');
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -552,7 +552,7 @@ This program is available under Apache License Version 2.0, available at https:/
 
     /** @private */
     _closeOrCommit() {
-      if (this.autoOpenDisabled && !this.opened) {
+      if (this.autoOpenDisabled && !this.opened && !this.loading) {
         this._commitValue();
       } else {
         this.close();
@@ -685,9 +685,12 @@ This program is available under Apache License Version 2.0, available at https:/
           this.value = '';
         }
       } else {
+        const itemsMatchedByLabel = this.filteredItems
+          && this.filteredItems.filter(item => this._getItemLabel(item) === this._inputElementValue)
+          || [];
         if (this.allowCustomValue
           // to prevent a repetitive input value being saved after pressing ESC and Tab.
-          && !(this.filteredItems && this.filteredItems.filter(item => this._getItemLabel(item) === this._inputElementValue).length)) {
+          && !itemsMatchedByLabel.length) {
 
           const e = new CustomEvent('custom-value-set', {detail: this._inputElementValue, composed: true, cancelable: true, bubbles: true});
           this.dispatchEvent(e);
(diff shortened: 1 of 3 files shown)


src/vaadin-combo-box-mixin.html
```diff
@@ -696,6 +699,8 @@
             this._selectItemForValue(customValue);
             this.value = customValue;
           }
+        } else if (!this.allowCustomValue && !this.opened && itemsMatchedByLabel.length == 1) {
```
yuriy-fix, line 702: We probably should test how it behaves with `allowCustomValue`. 🤔

src/vaadin-combo-box-mixin.html
```diff
@@ -685,9 +685,12 @@
           this.value = '';
         }
       } else {
+        const itemsMatchedByLabel = this.filteredItems
```
yuriy-fix, line 688: Didn't understand it from the ticket specifications. Should we have one **exact** match or should it be **one** match? For example, "Finlan" -> "Finland".

src/vaadin-combo-box-mixin.html
```diff
@@ -552,7 +552,7 @@
 
     /** @private */
     _closeOrCommit() {
-      if (this.autoOpenDisabled && !this.opened) {
+      if (this.autoOpenDisabled && !this.opened && !this.loading) {
```
yuriy-fix, line 555: Nit: not related to this PR, but `loading` could be a bit dangerous to rely on as it persists after leaving the field with pending requests **and is not removed afterwards.**
loading
**Definitely not a blocker**

test/lazy-loading.html
```diff
@@ -852,6 +852,20 @@
               const pages = spyDataProvider.getCalls().map(call => call.args[0].page);
               expect(pages).to.contain(1);
             });
+
+            it('should set value if not opened nor focused', () => {
+              const dataProvider = (params, callback) => {
+                comboBox.blur();
```
yuriy-fix, line 858: Nit: can we move `blurring` out of the `dataProvider`? I guess it's quite a rare case to `blur` the field from DP.

yuriy-fix (commented):
This change checks for `autoOpenDisabled` and not applying the value if `autoOpenDisabled` is off.
I guess it should be changed to follow Diego's [comment](https://github.com/vaadin/vaadin-combo-box/issues/900#issuecomment-698350156) which specifies that it should behave regardless if `autoOpenDisabled` is on or off.

test/lazy-loading.html
tulioag: The intention is that the field is focused when  `dataProvider()` is called, but is not focused when the results return and `callback()` is called. The tests fail otherwise

src/vaadin-combo-box-mixin.html
tulioag: My understating is that is should be **exact**.  Setting on partial matching would be a new feature. See https://github.com/vaadin/vaadin-combo-box/issues/775#issuecomment-501979966

src/vaadin-combo-box-mixin.html
tulioag: In this example, has the data provider returned the data? Data provider should set loading to false on return.

```javascript
            if (Object.keys(this._pendingRequests).length === 0) {
              this.loading = false;
            }
```

tulioag: https://github.com/vaadin/vaadin-combo-box/blob/5007fcc182ed59fb1813e79e24349612561a781b/src/vaadin-combo-box-data-provider-mixin.html#L168-L170

src/vaadin-combo-box-mixin.html
tulioag: If I remove the `!allowCustomValue`, this test fails:
https://github.com/vaadin/vaadin-combo-box/blob/5007fcc182ed59fb1813e79e24349612561a781b/test/vaadin-combo-box.html#L204-L220

combobox_custom_value_test_error

tulioag:
> This change checks for `autoOpenDisabled` and not applying the value if `autoOpenDisabled` is off.
> I guess it should be changed to follow Diego's [comment](https://github.com/vaadin/vaadin-combo-box/issues/900#issuecomment-698350156) which specifies that it should behave regardless if `autoOpenDisabled` is on or off.

The behaviour is consistent in both cases, since when data is received `this._commitValue` is called regardless of `autoOpenDisabled`'s value. Added tests covering both cases.

src/vaadin-combo-box-mixin.html
yuriy-fix: This is happening on the `Lazy loading` in the `Basic` if you quickly type the value and leave the field. It doesn't pass the first [condition](https://github.com/vaadin/vaadin-combo-box/blob/5007fcc182ed59fb1813e79e24349612561a781b/src/vaadin-combo-box-data-provider-mixin.html#L152) as there is no pending requests anymore. 
But as said this is happening on master as well, so not caused by this PR.

test/lazy-loading.html
yuriy-fix: I understand the intention of the test. I do mean that either the description of the test can be updated (i.e. `if not opened nor focused ` + **anymore**) or blur can be called outside the dp or the comment can be added to the line explaining why it is there 😅

yuriy-fix:
> 
> The behaviour is consistent in both cases, since when data is received `this._commitValue` is called regardless of `autoOpenDisabled`'s value. Added tests covering both cases.

Have tried on the demo page. The valid value `Finland` is cleared right away on blurring the field if `autoOpenDisabled` is not applied.
aod

src/vaadin-combo-box-mixin.html
tulioag: Created issue #963

yuriy-fix (approved):

Issue #970, opened 2020-11-16
Scrolling quickly to the bottom of the dropdown shows blank items


johannesh2:
### Description
Scrolling a lazy loading combo box to the bottom of the result sets shows blank dropdown. Scrolling back up a bit, reveals the items.

### Code
```java
import java.util.ArrayList;
import java.util.Optional;
import java.util.stream.Stream;

import com.vaadin.flow.component.combobox.ComboBox;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.data.provider.Query;
import com.vaadin.flow.router.Route;

@Route("lazycombo")
public class LazyCombo extends VerticalLayout {
  static final ArrayList<String> list = new ArrayList<>();
  static final int MAX_ITEMS = 500;
  static {
    for (int i = 1; i <= MAX_ITEMS; i++) {
      list.add(String.valueOf(i));
    }
  }
  
  public LazyCombo() {
    ComboBox<String> combo = new ComboBox<>();
    combo.setItems(this::items);
    add(combo);
  }
  
  private int count(Query<String, String> q) {
    int count = (int) filteredItems(q).count();
    return count;
  
  }
  
  private Stream<String> items(Query<String, String> q) {
    return filteredItems(q).skip(q.getOffset()).limit(q.getLimit());
  }
  
  private Stream<String> filteredItems(Query<String, String> q) {
    return list.stream().filter(item -> doFilter(item, q.getFilter()));
  }
  
  private boolean doFilter(String item, Optional<String> filter) {      
    if (filter.isEmpty()) {
      return true;
    } else {
      return item.contains(filter.get()); 
    }
  }
}
```

### Gif
![comboscrollbug](https://user-images.githubusercontent.com/1398470/99222635-a4c36a00-27eb-11eb-8b66-e82f5fd62ffc.gif)

Tested with V18 beta2. On Win 10 Chrome 86

mshabarov:
This happens when you drag and scroll sharply. Apparently, 2-3 last pages are not requested from backend and, thus, rendered.

This video shows two cases: with sharp scrolling and with smooth scrolling. All the pages are fetched in second case.

https://drive.google.com/file/d/1EtyuPIXFifKKuawuCslORf9aspp5pFBP/view?usp=sharing.

Proceed with investigation.

PR #978, opened 2020-12-02
fix: Scrolling quickly to the bottom shows blank items

mshabarov:
When the dropdown was being scrolled quickly, some pages weren't being loaded and skipped in the end. It's fixed by resending the skipped pages after the scroller reset and filter those indexes which are close to current scroller position, so as to avoid unnecessary duplicated pages and loading indicator hanging.

Fixes: https://github.com/vaadin/vaadin-combo-box/issues/970

In draft still, because need to cover these changes by a tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 dist: xenial
 language: node_js
-node_js: 8.11
+node_js: 10.23
 
 cache:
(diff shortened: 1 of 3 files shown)


tomivirkki (approved):
The hotfix LGTM, we agreed to add the tests afterward.

PR #987, opened 2021-01-05
feat!: support generics for item type

### LINKED ISSUE #985: TypeScript generics to allow declaring GridItem type
### Description

See https://github.com/vaadin/vaadin-grid/issues/2104 for more context.

The implementation has been prototyped for `vaadin-crud`, see https://github.com/vaadin/vaadin-crud/pull/235/commits/e2f774733ccb2083165227a2429c7eeee2706ced

The idea is to allow passing generic type for `ComboBoxElement` and then pass it to the event map, same as for `vaadin-crud`.
The only difference is that in `vaadin-combo-box` the generic type needs to be also passed to a mixin:
https://github.com/vaadin/vaadin-combo-box/blob/fdac1e843a907d0f6f4b89c6c60dfe749452c606/src/vaadin-combo-box-data-provider-mixin.d.ts#L36

web-padawan:
Fixes #985 

This is a breaking change which consists of the following:

1. Removes `ComboBoxItem` which is just an alias for `unknown` and therefore is pretty much useless;
2. Change certain types to use generics using the type cast like this:

```ts
const comboBox = this.shadowRoot.querySelector('vaadin-combo-box') as ComboBoxElement<User>
````

Note, I had to copy-paste and override the related properties on the element classes instead of modifying relevant mixins. 
This is needed because of the TypeScript limitation, see https://github.com/vaadin/vaadin-combo-box/issues/985#issuecomment-749380363 and https://github.com/microsoft/TypeScript/issues/24122.

### DIFF
--- a/src/vaadin-combo-box-mixin.d.ts
+++ b/src/vaadin-combo-box-mixin.d.ts
@@ -1,4 +1,4 @@
-import { ComboBoxItem, ComboBoxRenderer } from './interfaces';
+import { ComboBoxRenderer } from './interfaces';
 
 declare function ComboBoxMixin<T extends new (...args: any[]) => {}>(base: T): T & ComboBoxMixinConstructor;
 
@@ -41,13 +41,13 @@ interface ComboBoxMixin {
    *   - `model.index` The index of the rendered item.
    *   - `model.item` The item.
    */
-  renderer: ComboBoxRenderer | null | undefined;
+  renderer: ComboBoxRenderer<any> | null | undefined;
 
   /**
    * A full set of items to filter the visible options from.
    * The items can be of either `String` or `Object` type.
    */
-  items: Array<ComboBoxItem | string> | undefined;
+  items: Array<unknown> | undefined;
 
   /**
    * If `true`, the user can input a value that is not present in the items list.
@@ -63,7 +63,7 @@ interface ComboBoxMixin {
    * can be assigned directly to omit the internal filtering functionality.
    * The items can be of either `String` or `Object` type.
    */
-  filteredItems: Array<ComboBoxItem | string> | undefined;
+  filteredItems: Array<unknown> | undefined;
 
   /**
    * The `String` value for the selected item of the combo box.
@@ -90,7 +90,7 @@ interface ComboBoxMixin {
   /**
    * The selected item from the `items` array.
    */
-  selectedItem: ComboBoxItem | string | null | undefined;
+  selectedItem: unknown | null | undefined;
 
   /**
    * Path for label of the item. If `items` is an array of objects, the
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/vaadin-combo-box.d.ts [hunk 1]
```diff
@@ -237,22 +283,22 @@ declare class ComboBoxElement extends ElementMixin(
...
-    listener: (this: ComboBoxElement, ev: ComboBoxEventMap[K]) => void,
+    listener: (this: ComboBoxElement<Item>, ev: ComboBoxEventMap<Item>[K]) => void,
     options?: boolean | EventListenerOptions
   ): void;
 }
 
 declare global {
   interface HTMLElementTagNameMap {
-    'vaadin-combo-box': ComboBoxElement;
+    'vaadin-combo-box': ComboBoxElement<any>;
```
vlukashov, line 301: Could you please explain what's the main reason for using `any` instead of `unknown` here? Maybe add a comment.

It's not obvious when why some places use `any` (here, and e.g. `renderer: ComboBoxRenderer<any> | null | undefined;`) while some other places use `unknown` (e.g. `items: Array<unknown> | undefined;` and `dataProvider: ComboBoxDataProvider<unknown> | null | undefined;`).

src/vaadin-combo-box-mixin.d.ts
```diff
@@ -63,7 +63,7 @@ interface ComboBoxMixin {
    * can be assigned directly to omit the internal filtering functionality.
    * The items can be of either `String` or `Object` type.
```
vlukashov, line 64: > The items can be of either `String` or `Object` type.
This is now inconsistent with the type definitions, is not it? Should we update the comment or the type definitions?

src/vaadin-combo-box.d.ts
web-padawan: I had to use `any` to get rid of the  following error:

```
  Type 'ComboBoxRenderer<Item> | null | undefined' is not assignable to type 'ComboBoxRenderer<unknown> | null | undefined'.
    Type 'ComboBoxRenderer<Item>' is not assignable to type 'ComboBoxRenderer<unknown>'.
      Type 'unknown' is not assignable to type 'Item'.
        'Item' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.
```

Added a comment above explaining why `any` is used for the mixin interface.

src/vaadin-combo-box-mixin.d.ts
web-padawan: I could not get TypeScript to work properly using `string | object` neither `string | Record<string, unknown>`.

IMO we could keep the comment as it describes the way component is designed to work.

web-padawan:
Closing this for now because of https://github.com/microsoft/TypeScript/issues/24122, will try to find a better solution later.

PR #1001, opened 2021-03-03
fix: filter when opening and auto-open is disabled

### LINKED ISSUE #1000: No filter is applied on a lazy-loading, auto-open-disabled combo-box, when opening it
### Description

For a combo-box with lazy-loading, when `auto-open-disabled` is set to `true`, once the user writes anything in the combo-box, clicking the button to open the combo-box will bring the complete list of items, disregarding the user's text and not filtering the results.

Once opened, the filter works as expected.

### Live Demo

https://user-images.githubusercontent.com/76832183/108853339-46c6cf80-75ef-11eb-9c62-c9b48aad8e88.mov

### Steps to reproduce
The bug can be reproduced in the demo view in it's flow counterpart. Just add the following code above [this line](https://github.com/vaadin/vaadin-flow-components/blob/a5ccd8d9b8f00003359e8722d5b03747ac6d60af/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-demo/src/main/java/com/vaadin/flow/component/combobox/demo/ComboBoxView.java#L345).

```java
comboBox.setAutoOpen(false);
```

For the web component version:
```html
<vaadin-combo-box label="Country" auto-open-disabled="true"></vaadin-combo-box>
```
```javascript
customElements.whenDefined('vaadin-combo-box').then(() => {
    const comboBox = document.querySelector('vaadin-combo-box');
    comboBox.dataProvider = function(params, callback) {
        const xhr = new XMLHttpRequest();
        xhr.onload = function() {
          const response = JSON.parse(xhr.responseText);
          callback(response.result, response.size);
        };
        const index = params.page * params.pageSize;
        const url = 'https://demo.vaadin.com/demo-data/1.0/filtered-countries?index=' + index +
          '&count=' + params.pageSize +
          '&filter=' + params.filter;
        xhr.open('GET', url, true);
        xhr.send();
      };
});
```

### Expected Results
The list should be filtered right after opening the combo-box.

### Actual Results
The complete list is rendered when opened, no filtering applied.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] iOS Safari <- not tested
- [ ] Android Chrome <- not tested

### Version

- vaadin-combo-box: v6.0.0

sosa-vaadin:
Fixes #1000

### DIFF
--- a/src/vaadin-combo-box-data-provider-mixin.js
+++ b/src/vaadin-combo-box-data-provider-mixin.js
@@ -75,10 +75,10 @@ export const ComboBoxDataProviderMixin = (superClass) =>
     }
 
     /** @private */
-    _dataProviderClearFilter(dataProvider) {
+    _dataProviderClearFilter(dataProvider, opened, value) {
       // Can't depend on filter in this observer as we don't want
       // to clear the filter whenever it's set
-      if (dataProvider && !this.loading && this.filter) {
+      if (dataProvider && !this.loading && this.filter && !(opened && this.autoOpenDisabled && value === this.filter)) {
         this.size = undefined;
         this._pendingRequests = {};
         this.filter = '';
--- a/test/lazy-loading.test.js
+++ b/test/lazy-loading.test.js
@@ -123,6 +123,23 @@ describe('lazy loading', () => {
         expect(comboBox.selectedItem).to.be.null;
       });
 
+      (isComboBoxLight ? describe.skip : describe)('when autoOpenDisabled', () => {
+        beforeEach(() => (comboBox.autoOpenDisabled = true));
+
+        it('should not clear filter when loading first page', () => {
+          expect(comboBox.autoOpenDisabled).to.be.true;
+          expect(comboBox.opened).to.be.false;
+          comboBox.dataProvider = spyDataProvider;
+          setInputValue('item 1');
+          comboBox.inputElement.dispatchEvent(new CustomEvent('input'));
+          comboBox.opened = true;
+          flush();
+          expect(comboBox.filter).to.equal('item 1');
+          const { filter } = spyDataProvider.lastCall.args[0];
+          expect(filter).to.equal('item 1');
+        });
+      });
+
       describe('when open', function () {
         this.timeout(15000);
         beforeEach(() => (comboBox.opened = true));
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@vaadin/vaadin-element-mixin": "^2.4.1"
   },
   "devDependencies": {
-    "@esm-bundle/chai": "^4.1.5",
+    "@esm-bundle/chai": "4.3.0",
     "@open-wc/rollup-plugin-html": "^1.2.5",
     "@open-wc/testing-helpers": "^1.8.0",
     "@polymer/iron-component-page": "^4.0.0",


src/vaadin-combo-box-data-provider-mixin.js
```diff
@@ -169,6 +169,12 @@ export const ComboBoxDataProviderMixin = (superClass) =>
       if (!this._pendingRequests[page] && this.dataProvider) {
         this.loading = true;
 
+        // do not clear filter if auto open is disabled
+        // otherwise, the first page loaded will not be filtered
+        if (this.autoOpenDisabled) {
+          this.filter = this.filter || this._inputElementValue;
+        }
```
web-padawan, line 176: I would suggest to go with a slightly different approach to avoid extra `clearCache` call. 

So, here is a problem:

1. When combo-box has `autoOpenDisabled`, and user types a value, both `value` and `filter` are set.
2. The following observer is called because `value` is set before opening the combo-box:
https://github.com/vaadin/vaadin-combo-box/blob/f7fdbe508adfca584216cf0578c922fbfc8d39c5/src/vaadin-combo-box-data-provider-mixin.js#L81-L86
3. `this.filter` check returns `true` and logic inside `id` is executed, so it resets the `filter`, which is a bug.

Proposed fix: update the observer condition to look like this:

```js
if (dataProvider && !this.loading && this.filter && !(this.autoOpenDisabled && value === this.filter)) {
```

That's exactly what happens: `value` and `filter` are the same, and `autoOpenDisabled` is true.

test/lazy-loading.test.js
```diff
@@ -123,6 +123,22 @@ describe('lazy loading', () => {
...
       });
 
+      describe('when autoOpenDisabled', () => {
+        beforeEach(() => (comboBox.autoOpenDisabled = true));
+
+        (isComboBoxLight ? it.skip : it)('should not clear filter when loading first page', () => {
+          expect(comboBox.autoOpenDisabled).to.be.true;
+          expect(comboBox.opened).to.be.false;
+          comboBox.dataProvider = spyDataProvider;
+          comboBox.filter = 'item 1';
```
web-padawan, line 133: Let's mimic the user input instead to emulate setting `filter`:

```suggestion
          setInputValue('item 1');
          comboBox.inputElement.dispatchEvent(new CustomEvent('input'));
```

test/lazy-loading.test.js
```diff
@@ -123,6 +123,22 @@ describe('lazy loading', () => {
         expect(comboBox.selectedItem).to.be.null;
       });
 
+      describe('when autoOpenDisabled', () => {
+        beforeEach(() => (comboBox.autoOpenDisabled = true));
+
+        (isComboBoxLight ? it.skip : it)('should not clear filter when loading first page', () => {
```
web-padawan, line 129: The test is passing also for `vaadin-combo-box-light` so let's not skip it.

```suggestion
        it('should not clear filter when loading first page', () => {
```

web-padawan (changes requested):
Thanks for investigating this bug. The test looks mostly good, added some minor comments.

But please check the solution that I proposed. It seems to be a bit more efficient.

sosa-vaadin:
@web-padawan  I was using `setInputValue()` but it was failing with `combo-box-light` which is why I was skipping the test. I changed it to setting the filter's value directly because it was passing with both but forgot to change it back to run both tests. After playing around with the proposed changes, I ended up with these results:


  | setInputValue() | comboBox.filter
-- | -- | --
Current fix | Combo-box-light fails | Both pass
Proposed fix | Both fail | Both pass

"Current fix" refers to the change as submitted in the PR. "Proposed fix" is the change in the condition you proposed.

The reason it fails with the current fix for `combo-box-light` is because of how the `setInputValue()` function is implemented:

```javascript
  const setInputValue = (value) => {
    if (comboBox.inputElement.tagName === 'IRON-INPUT') {
      comboBox.inputElement._initSlottedInput();
      comboBox.inputElement.inputElement.value = value;
    } else {
      comboBox.inputElement.value = value;
    }
  };
```
The `if` condition resolves to `true` for `combo-box-light` and so

```javascript
if (this.autoOpenDisabled) {
  this.filter = this.filter || this._inputElementValue;
}
```
`this._inputElementValue` will have an empty string as value, since it was not set with `setInputValue()`

web-padawan:
@sosa-vaadin You are correct, the proposed fix also fails with `vaadin-combo-box-light`.

Please see my commit at https://github.com/vaadin/vaadin-combo-box/commit/4bfcabe00c19c19872257edba54c77d9ac1bfb51, for the `vaadin-combo-box` it passes.

test/lazy-loading.test.js
web-padawan: Actually, it fails.

web-padawan (approved):
LGTM, rebased to make the first commit contain a final version of a fix.

Will merge this after the tests pass.

PR #1006, opened 2021-05-10
fix: Make server-side filtering case-insensitive

tulioag:
Fixes: vaadin/flow-components#915

### DIFF
--- a/test/lazy-loading.html
+++ b/test/lazy-loading.html
@@ -976,8 +976,9 @@
               expect(comboBox.value).to.equal('other value');
 
               returnedItems = ['item 12'];
-              comboBox._inputElementValue = 'item 1';
-              comboBox.filter = 'item 1';
+              const filterValue = 'item 1';
+              comboBox._inputElementValue = filterValue;
+              comboBox.filter = filterValue;
               expect(comboBox.opened).to.be.false;
               expect(comboBox.hasAttribute('focused')).to.be.false;
               expect(comboBox.value).to.equal('other value');
--- a/test/lazy-loading.html
+++ b/test/lazy-loading.html
@@ -907,8 +907,9 @@
               comboBox.autoOpenDisabled = false;
               expect(comboBox.autoOpenDisabled).to.be.false;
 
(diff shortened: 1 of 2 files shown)


test/lazy-loading.html
```diff
@@ -925,6 +925,38 @@
...
+              expect(comboBox.autoOpenDisabled).to.be.true;
+
+              comboBox._inputElementValue = 'iTem 12';
+              comboBox.filter = 'item 12';
+              expect(comboBox.opened).to.be.false;
+              expect(comboBox.hasAttribute('focused')).to.be.false;
+              expect(comboBox.value).to.equal('item 12');
+            });
+
+            it('should set value with multiple matches that differ only in case', () => {
```
sissbruecker, line 950: If I understand it correctly, this test verifies that the first item is picked if there are multiple matches. I think the test name should reflect that.

test/lazy-loading.html
```diff
@@ -925,6 +925,38 @@
               expect(comboBox.value).to.equal('item 12');
             });
 
+            it('should set value without auto-open-disabled even if case does not match', () => {
+              comboBox.autoOpenDisabled = false;
+              expect(comboBox.autoOpenDisabled).to.be.false;
+
+              comboBox._inputElementValue = 'ItEm 12';
+              comboBox.filter = 'item 12';
```
sissbruecker, line 933: To help me understand this: `filter` is the value that the last data provider query was made for, and `_inputElementValue` is the current state of the input?

sissbruecker (changes requested):

test/lazy-loading.html
tulioag: Thanks for pointing it out. This was wrong. They both should have the same value, as `filter` gets the value of `_inputElementValue` on the input event.

test/lazy-loading.html
sissbruecker: @tulioag What about this one?

tulioag: Done

sissbruecker (approved):

PR #1010, opened 2021-07-28
CP: fix: do not clear filter when autoOpenDisabled

sosa-vaadin:
Cherry-pick of: https://github.com/vaadin/vaadin-combo-box/pull/1001


yuriy-fix:
The original fix is not aiming critical issue, so no need to backport it.

PR #1012, opened 2021-10-04
fix: do not fire custom-value-set event when readonly (CP: 14)

sissbruecker:
cherry-pick of: https://github.com/vaadin/web-components/pull/2721
related to: https://github.com/vaadin/flow-components/issues/2135

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -1050,7 +1050,7 @@ This program is available under Apache License Version 2.0, available at https:/
         event.composedPath()[0].focus();
         return;
       }
-      if (!this._closeOnBlurIsPrevented) {
+      if (!this.readonly && !this._closeOnBlurIsPrevented) {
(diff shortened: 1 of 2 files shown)


hdamr (commented):
I think it is better to move newly added test to this section, otherwise, it looks good to me.
https://github.com/vaadin/vaadin-combo-box/blob/a44551c9ea97c8c652bb2b06e946fd573d17ea01/test/vaadin-combo-box.html#L190

hdamr (approved):

sissbruecker:
> I think it is better to move newly added test to this section, otherwise, it looks good to me.
> 
> https://github.com/vaadin/vaadin-combo-box/blob/a44551c9ea97c8c652bb2b06e946fd573d17ea01/test/vaadin-combo-box.html#L190

Tried this initially, but the whole outer test suite uses some setup with which the issue is actually not reproducable. Instead of refactoring the whole file I decided to just add a new suite at the bottom.

PR #1015, opened 2021-11-05
fix: close vaadin-dialog on Esc in vaadin-combo-box

sosa-vaadin:
## Description

Pressing the <kbd>Esc</kbd> key on a closed `vaadin-combo-box` placed in a `vaadin-dialog` now closes the dialog.

Fixes https://github.com/vaadin/web-components/issues/984

## Type of change

- [x] Bugfix

### DIFF
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -580,16 +580,32 @@ This program is available under Apache License Version 2.0, available at https:/
      */
     _onEscape(e) {
       if (this.autoOpenDisabled) {
-        this._focusedIndex = -1;
-        this.cancel();
-      } else if (this.opened) {
-        this._stopPropagation(e);
-
-        if (this._focusedIndex > -1) {
+        // Auto-open is disabled
+        if (this.value !== this._inputElementValue && this._inputElementValue.length > 0) {
+          // The input value has changed but the change hasn't been committed, so cancel it.
           this._focusedIndex = -1;
-          this._revertInputValue();
-        } else {
           this.cancel();
+        } else if (this.clearButtonVisible && !this.opened) {
+          // The clear button is visible and the overlay is closed, so clear the value.
+          this._clear();
+        }
+      } else {
+        // Auto-open is enabled
+        if (this.opened) {
+          // The overlay is open
+          e.stopPropagation();
+
+          if (this._focusedIndex > -1) {
+            // An item is focused, revert the input to the filtered value
+            this._focusedIndex = -1;
+            this._revertInputValue();
+          } else {
+            // No item is focused, cancel the change and close the overlay
+            this.cancel();
+          }
+        } else if (this.clearButtonVisible) {
+          // The clear button is visible and the overlay is closed, so clear the value.
+          this._clear();
         }
       }
     }
--- a/src/vaadin-combo-box-mixin.html
+++ b/src/vaadin-combo-box-mixin.html
@@ -607,7 +607,7 @@ This program is available under Apache License Version 2.0, available at https:/
             this.cancel();
           }
         } else if (this.clearButtonVisible && !!this.value) {
-          // this._stopPropagation(e);
+          this._stopPropagation(e);
           // The clear button is visible and the overlay is closed, so clear the value.
           this._clear();
         }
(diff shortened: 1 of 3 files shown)


test/keyboard.html
```diff
@@ -284,11 +284,12 @@
       });
 
       it('escape key event should not be propagated', () => {
```
sissbruecker, line 286: Let's add a test to verify that the escape key event gets propagated if the dropdown is closed.

sissbruecker (changes requested):

sosa-vaadin:
Depends on https://github.com/vaadin/vaadin-text-field/pull/585

src/vaadin-combo-box-mixin.html
```diff
@@ -580,16 +580,32 @@
      */
     _onEscape(e) {
       if (this.autoOpenDisabled) {
```
sissbruecker, line 582: I think the implementation here has several issues:

- Escape is propagated when auto open is disabled. Don't understand why this branching exists. As long as the dropdown is open, then pressing Escape should only close the dropdown, but not propagate the event further.
- When the clear button is visible, then clearing the text field value with Escape also causes the event to be propagated. Clearing the text field value should not cause dialogs to close.

src/vaadin-combo-box.html
```diff
@@ -397,9 +397,8 @@
           // We need to set this listener for "this.inputElement"
           // instead of just "this", otherwise keyboard navigation behaviour
           // breaks a bit on Safari and some related tests fail.
-          this.inputElement.addEventListener('keydown', e => {
+          this.addEventListener('keydown', e => {
```
tomivirkki, line 400: There are 2 "keydown" listeners for the `<vaadin-text-field>`, one in text-field (handles clear button) and another one here in combo-box. This one is expected to stop the propagation of the event under certain conditions so it doesn't get processed by the one in text-field (and doesn't get a dialog closed either).

combo-box is listening for the event in capture phase, but in one of the tests on Firefox, that didn't seem to matter; the text-field's listener would still get processed before this one. Moved the vaadin-combo-box' listener to the host element, so that it will for sure catch the event before it ever reaches the text-field's listener.

src/vaadin-combo-box.html
```diff
@@ -397,9 +397,8 @@
           // We need to set this listener for "this.inputElement"
           // instead of just "this", otherwise keyboard navigation behaviour
           // breaks a bit on Safari and some related tests fail.
```
tomivirkki, line 399: Need to remove this comment in case the change below is fine (tests seem to pass fine on Safari)

src/vaadin-combo-box-mixin.html
```diff
@@ -580,16 +580,35 @@
...
+          // The input value has changed but the change hasn't been committed, so cancel it.
+          this._stopPropagation(e);
           this._focusedIndex = -1;
-          this._revertInputValue();
-        } else {
           this.cancel();
+        } else if (this.clearButtonVisible && !this.opened && !!this.value) {
+          this._stopPropagation(e);
+          // The clear button is visible and the overlay is closed, so clear the value.
+          this._clear();
```
tomivirkki, line 592: The PR has a regression which should be covered in the `_onEscape` logic: if you open the dropdown of an `autoOpenDisabled` combobox, pressing esc now also clears the value

https://user-images.githubusercontent.com/1222264/141256666-b0ba2c40-c364-44dc-ba5d-b7a49e88937e.mp4

sosa-vaadin: Curious. I added a test for that specific scenario and it's passing. I must check it.

sosa-vaadin: Ah, the value of the combo-box doesn't change but the value of the internal text-field does.

sosa-vaadin: Added an extra condition to the first check on autoOpenDisabled instances.

src/vaadin-combo-box-mixin.html
sissbruecker: All of those cases are resolved now.

sissbruecker (approved):

PR #1052, opened 2023-09-21
fix: disable validation of internal text-field

vursen:
## Description

The PR disables the internal text-field's validation so that it doesn't override the propagated invalid state.

Fixes https://github.com/vaadin/flow-components/issues/5409

> **Warning**
> Needs a cherry-pick to 14.10

## Type of change

- [x] Bugfix


vursen:
The tests failed in iOS Simulator due to a timeout. I think it's not a big deal, especially since they passed in all the other environments.

web-padawan (approved):

